PR #282, opened 2026-08-16
Block committed agent-local permission files

sarthak-here:
## Related issue

N/A. Raised with @elCaptnCode directly and scoped there.

## Summary

`.claude/settings.local.json` was tracked in the repo. It is a local Claude Code
permission file, and it pre-approved tool use such as `Bash(curl:*)`,
`Bash(powershell -Command:*)`, `Bash(taskkill:*)`, and
`Bash(Remove-Item -Recurse -Force "...")`. Anyone who cloned the repo and opened
it in Claude Code inherited those approvals with no prompt at the point of use.

Nothing guarded this. There was no `.gitignore` entry and no governance check.

This PR:

- untracks `.claude/settings.local.json` and removes it from the working tree
- gitignores agent-local settings for `.claude`, `.agents`, and `.codex`
- adds an `agent_local_config_committed` error to
  `scripts/governance_guardrails.py` so it cannot come back
- adds two tests, one for the error and one for the scope boundary

### Scope

The guardrail matches an explicit path set, not `.claude/**`:

```python
AGENT_LOCAL_CONFIG_PATHS = {
    ".agents/settings.local.json",
    ".claude/settings.local.json",
    ".codex/settings.local.json",
}
```

Matching is exact set membership, so shared project guidance under
`.claude/rules/` and `.claude/skills/` cannot be caught by it. All 37 tracked
files under `.claude/` were scanned after the change and produced zero findings.
`test_governance_guardrail_keeps_shared_agent_guidance_tracked` locks that
boundary in, so widening the rule to `.claude/**` later fails a test.

`.agents` and `.codex` are included because they are the same class of file and
`.claude/rules/multi-agent-coordination.md` describes Codex working in this
repo. Happy to narrow to Claude only if you prefer.

### One tradeoff worth a look

The check is a pure path match with no git call, which keeps it deterministic
and easy to test. The consequence is that `_iter_all_files` walks the working
tree rather than the index, so a contributor who has their own correctly
gitignored `settings.local.json` and runs `--all` locally will see the error.
Default mode is unaffected, since `_iter_changed_files` uses
`--exclude-standard`.

Making `--all` respect gitignore would fix that, but it changes scan semantics
repo wide, which felt outside the scope set for this change. Happy to do it as a
follow up.

## Tests run

```
python -m pytest tests/test_governance_guardrails.py -q --no-cov  -> 9 passed
```

That includes the existing `test_governance_guardrail_passes_current_repo_all_scan`,
which asserts a clean repo-wide scan, so it confirms the repo itself is now clean
rather than only guarded.

End to end against the real script:

```
# with the file restored temporarily
python scripts/governance_guardrails.py --all --errors-only
ERROR agent_local_config_committed .claude/settings.local.json: agent-local
permission/config files must stay untracked; remove from git tracking and add
the path to .gitignore
Governance guardrails failed with 1 error(s).
exit=1

# with the file removed
python scripts/governance_guardrails.py --all --errors-only
Governance guardrails passed.
exit=0
```

I could not run `ruff` locally, the binary was not available in my environment,
so CI is the first lint pass on this.

## OSS Change Impact

- Layer changed: repo governance tooling only
- Build workflow impact: none
- Runtime/platform impact: none
- App workspace impact: none
- Tests run: see above
- Docs updated: none required
- Compatibility risk: low

## Boundary confirmation

- [x] This change stays within the open-source `mozaiks` repository and does not introduce private hosted-product logic, credentials, or internal-only URLs.

## Governance check

- [x] This is an ordinary change with no new public schema, public workflow/prompt family, provider mutation path, authority bypass, eval artifact, learned optimization, or cross-customer intelligence.
- [x] No public contract is added or changed, so classification and versioning do not apply.
- [x] This does not touch a one-way-door area from `OSS_PUBLICATION_POLICY.md`, so no ADR is included.
- [x] This does touch permissions, in the sense of contributor-local agent tool
      approvals. The authority path is explicit: agent-local permission files
      become untracked and a governance error blocks them from returning. No
      runtime, module, or `granted_permissions` authority behavior changes.


elCaptnCode:
This looks like the right fix. Scope is correct:

- removes tracked \.claude/settings.local.json\;
- ignores exact agent-local settings files;
- adds a governance guardrail for committed local permission/config files;
- preserves shared \.claude/rules/**\ and \.claude/skills/**\.

Please keep this as a small standalone security/governance PR, wait for the normal required checks to finish, then squash-merge if green.

sarthak-here:
checks are all green, 9/9 including lint and the full test suite. can't merge it myself though, i only have read access on the repo. all yours whenever

elCaptnCode:
Superseded by #285 — your contribution is included with full author credit. The correction fixes the working-tree vs tracked-file distinction using git's index authority. Thank you for the original finding and fix!

PR #295, opened 2026-08-18
 feat: add lead qualifier AI agent workflow and templates 

laibapafaftab-debug:
Added a sample B2B Lead Qualifier AI Agent module (`app/modules/lead_qualifier`) using Jinja prompt templates and structured JSON schema for monetizable SaaS workflows.


sarthak-here:
Thanks for putting this up. It needs some work before it can go in.

The big one: 81 of the 86 files are your local mozaiks-workspace/ directory. That's generated quickstart output, not repo content, and it includes duplicate copies of CLAUDE.md, AGENTS.md and the .claude/rules and .claude/skills files. That's on us for not gitignoring it, I'll fix that separately, but they need dropping from the branch.

refinement_policy.yaml loses 26 lines, all of it documentation and the commented-out openai profile options. Comments stripped and long strings reflowed usually means a YAML load-and-dump ran over the file. Worth reverting that file entirely, it isn't related to the module.

The ai.json addition won't take effect. The runtime reads api_type, not provider, and nothing reads an llm block out of ai.json.

On the module itself, the path is app/modules/app/modules/lead_qualifier/ with app/modules doubled, and it only has prompt.jinja and schema.json. A loadable module needs module.yaml with identity and actions, plus backend/handler.py. As it stands the runtime can't pick it up.

The prompt is good, clear role and clean variables. Suggest a fresh branch off main with just the module files, laid out to the module contract.

elCaptnCode:
so this is an app built with mozaiks i wouldn't consider this as a PR.  i think we should create maybe a whole different repo with test apps or something.  

also we are not looking to create agents quite yet. (and if we did we would use the conical structure of an [agent ](https://docs.mozaiks.ai/guides/adding-workflows/01-overview/) and put it under [app/build_context](https://docs.mozaiks.ai/architecture/workflows/build-context-packs/?h=build_conte#build-context-vs-context-variables) library eventually.

but i dont think we are quiet there yet. what we are looking for is help hardening the current factory-app which generates the apps and agents. 

also mozaiks doesn't need prompts with jinja files if u see it has its own[ agnets.yaml](https://github.com/BlocUnited-LLC/mozaiks/pull/295/changes#diff-b50e9cdfea49ff3505a89f4776229a3c9bce237bcdd8002882d9f7ca7ce24a94) for the prompts for the agents it generates for you. but the conical structure built for your app is in factory_app/app/ and factory_app/app/workflows. 

you can start [here ](https://docs.mozaiks.ai/guides/studio/01-overview/ ) and [here ](https://docs.mozaiks.ai/architecture/app/canonical-app-structure/?h=app+structure) this describes how mozaiks itself is built.

elCaptnCode:
Closing PR

PR #302, opened 2026-08-19
Gitignore the generated quickstart workspace

sarthak-here:
## Summary

The README tells users to run `python -m mozaiks quickstart --dir .\mozaiks-workspace`, which generates a workspace at that path. Nothing in `.gitignore` covers it, so anyone following the Quickstart and running `git add -A` commits their whole generated workspace.

That workspace contains copies of repo files, including `CLAUDE.md`, `AGENTS.md`, `README.md`, the `.claude/rules` and `.claude/skills` files, and the `run-*.ps1` scripts. Committing those creates a second copy of project guidance that nobody can tell is non-authoritative.

This already happened: PR #295 carried 81 such files out of 86 changed.

## Change

One rule in `.gitignore`, plus two small tests.

The rule is scoped to `mozaiks-workspace/` only. Repo-owned files at the root stay tracked and editable, and one of the tests asserts that, so the rule cannot be widened later without failing.

## Tests run

```
python -m pytest tests/test_quickstart_workspace_hygiene.py -q --no-cov  ->  2 passed
```

## OSS Change Impact

- Layer changed: repo hygiene only
- Build workflow impact: none
- Runtime/platform impact: none
- App workspace impact: none
- Tests run: see above
- Docs updated: none required
- Compatibility risk: none. Nothing under `mozaiks-workspace/` is currently tracked, so no existing file becomes ignored.

## Boundary confirmation

- [x] This change stays within the open-source `mozaiks` repository and does not introduce private hosted-product logic, credentials, or internal-only URLs.

## Governance check

- [x] Ordinary change. No new public schema, workflow or prompt family, provider mutation path, authority bypass, eval artifact, learned optimization, or cross-customer intelligence.
- [x] No public contract added or changed.
- [x] Not a one-way-door area from `OSS_PUBLICATION_POLICY.md`, so no ADR.
- [x] Does not touch permissions, secrets, provider execution, payments, deployment, DNS, or production operations.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -122,3 +122,6 @@ web_shell/.mozaiks-tailwind-sources/
 .agents/settings.local.json
 .claude/settings.local.json
 .codex/settings.local.json
+
+# Local quickstart workspace (generated output, contains copies of repo files)
+mozaiks-workspace/
--- a/tests/test_quickstart_workspace_hygiene.py
+++ b/tests/test_quickstart_workspace_hygiene.py
@@ -1,0 +1,32 @@
+"""The quickstart workspace is generated output and must stay out of the repo."""
+
+from __future__ import annotations
+
+import subprocess
+from pathlib import Path
+
+REPO_ROOT = Path(__file__).resolve().parents[1]
+
+
+def _is_ignored(relative_path: str) -> bool:
+    result = subprocess.run(
+        ["git", "check-ignore", "--quiet", relative_path],
+        cwd=REPO_ROOT,
+        check=False,
+        capture_output=True,
+    )
+    return result.returncode == 0
+
+
+def test_quickstart_workspace_is_gitignored() -> None:
+    for relative_path in (
+        "mozaiks-workspace/CLAUDE.md",
+        "mozaiks-workspace/.claude/rules/runtime.md",
+        "mozaiks-workspace/scripts/run-frontend.ps1",
+    ):
+        assert _is_ignored(relative_path), f"{relative_path} must be gitignored"
+
+
+def test_repo_owned_files_stay_tracked() -> None:
+    for relative_path in ("CLAUDE.md", "README.md", ".claude/rules/runtime.md"):
+        assert not _is_ignored(relative_path), f"{relative_path} must stay tracked"


elCaptnCode:
I agree with the repo-hygiene problem this PR is fixing, but I want to make sure we're separating that from how Mozaiks-generated apps are intended to be iteratively developed.

Mozaiks already has a Refinement Engine/harness where coding agents such as Claude Code or Codex can act as bounded refinement workers. The harness owns classification, scope, routing, validation, and artifact lineage; the coding agent performs the actual scoped implementation work. That separation is intentional:

https://docs.mozaiks.ai/architecture/workflows/workflow-architecture/

Generated apps also have a canonical structure and contracts that need to be preserved during iterative development:

https://docs.mozaiks.ai/architecture/app/canonical-app-structure/

So I think ignoring `mozaiks-workspace/` here is correct. What I don't think we should do long term is copy the framework's own `CLAUDE.md`, `AGENTS.md`, and contributor guidance wholesale into generated apps.

Instead, should we formalize a separate generated-app coding-agent contract/skill pack?

That would give us one canonical set of app-development instructions that can be consumed in both cases:

1. By the Refinement Harness when it invokes Claude Code/Codex as a refinement worker.
2. By Claude Code/Codex directly when a developer clones an existing Mozaiks-generated app and iterates on it locally outside the harness.

For existing generated apps, we could then expose an idempotent CLI bootstrap/update command rather than requiring regeneration, e.g. `mozaiks dev init/update`, which installs or refreshes the appropriate app-development skills/instructions.

That seems cleaner to me than treating the copied framework guidance as part of the generated app. The framework instructions remain framework-specific, while the generated-app guidance teaches coding agents how to safely modify a Mozaiks app without breaking its canonical contracts.

The main thing I want to preserve is the existing architecture: AG2 owns workflow-local orchestration, Mozaiks owns refinement/builder-session orchestration, and coding agents perform bounded implementation work. Local coding-agent support should be another consumer of the app-development contract rather than a competing refinement architecture.

sarthak-here:
app/provenance.yaml already has the version handshake dev update would need. It records mozaiks_ref and a contracts: block, and is loaded by Studio and refinement tooling. Adding app_development: mozaiks.app_development.v1 there would let dev update compare versions and refresh only when stale, and it keeps the harness and a local coding agent reading the same declared version.

Second, add-module, add-page, add-capability, add-branding and add-feature under .claude/skills/ are already app-authoring rather than framework-internals. The pack may be less about writing new instructions and more about separating those from the framework-change skills.

No disagreement on the boundary. A local dev init pack should be a consumer of the app-development contract, not a second refinement path.

elCaptnCode:
One follow-up thought on where MCP could fit here, because I think this may be the cleanest way to support local iterative development without weakening the deterministic Mozaiks refinement model.

For a non-technical way to think about it: the Refinement Engine is the **decision-maker**, the coding agent is the **worker**, and MCP can be the **adapter/toolbox** that lets that worker understand and safely interact with a Mozaiks app.

In the managed Mozaiks flow, I still think the architecture should remain:

```text
User asks for a refinement
        ↓
Mozaiks Refinement Engine / Harness
        ↓
Decides what kind of change this is, what it may affect, and what scope is allowed
        ↓
Coding agent (Codex / Claude Code / other provider)
        ↓
Validation against canonical Mozaiks contracts
        ↓
New artifact version / preview / promotion
```

MCP would not replace that loop. It would sit closer to the coding-agent boundary and expose Mozaiks-specific capabilities in a standard way.

Conceptually:

```text
                   Canonical Mozaiks contracts/services
                 /          |          |           \
          inspect app   get structure  validate   bounded operations
                 \          |          |           /
                          Mozaiks MCP
                              ↓
                    Codex / Claude / Cursor
```

So instead of asking a coding agent to infer how a Mozaiks app works from a pile of copied framework files, the agent could ask Mozaiks directly through MCP things like:

- What type/version of Mozaiks workspace is this?
- What is the canonical structure for this app?
- Where should this new capability live?
- What modules/actions/contracts already exist?
- Which files are safe for this requested refinement to touch?
- Does this proposed change violate a Mozaiks contract?
- Validate the app after the edit.

That feels especially useful for the second use case: a developer clones an existing Mozaiks-generated app and opens it directly in Codex, Claude Code, Cursor, etc. They are no longer inside the managed Refinement Harness, but the coding agent could still connect to the same Mozaiks contract/validation capabilities through MCP and therefore continue developing the app according to its deterministic structure.

AG2 already has native MCP support, including consuming MCP servers and exposing agents over MCP, so this also seems directionally aligned with the existing AG2 ownership boundary rather than inventing a competing agent protocol:

https://docs.ag2.ai/docs/user-guide/tools/mcp_servers/
https://docs.ag2.ai/latest/docs/user-guide/cli/serve/

My current mental model is therefore:

```text
Managed refinement:
Mozaiks Harness → coding worker → Mozaiks capabilities

Local development:
Codex / Claude / Cursor → Mozaiks MCP → same Mozaiks capabilities
```

The important part is that the **canonical app contract and validation logic remain the source of truth**. MCP would just be an adapter that gives coding agents a portable, structured way to consume those capabilities.

If that model is sound, it may also mean the generated `AGENTS.md` / `CLAUDE.md` can stay very small: identify the project as a Mozaiks app, state the core invariants, and tell capable coding agents to use the Mozaiks MCP/tools for architecture-aware development. Static guidance would remain the fallback, while MCP provides the richer deterministic interface.

sarthak-here:
Split read from write. Read and validate tools are safe and give most of the value. A tool that can change app state becomes a new authority surface and a way around the harness. Ship read plus validate first.

Use the framework validator. There are three app_validation.py files. MCP should call mozaiksai/control_plane/, not the harness one, or local dev ends up depending on factory policy.

Provenance already answers "what workspace is this". app/provenance.yaml has mozaiks_ref and the contracts: block.

An MCP contract is a new public protocol surface, so it needs a short ADR under OSS_PUBLICATION_POLICY.md before implementation.

Agreed on the thin AGENTS.md. This PR is unrelated and green, so merge it as the hygiene fix and track MCP separately.

sarthak-here:
Opened #303 to track the ADR for the MCP contract surface and the read/write boundary, so the design discussion has a home and this PR can stay scoped to the gitignore fix.

Issue #309, opened 2026-08-19
test: add unit tests for `mozaiks info`


elCaptnCode:
## Problem

`mozaiks_cli/commands/info.py` (115 lines) has no dedicated test file. Compare `tests/` — there are focused tests for `quickstart`, `onboard`, `serve`, `migrations`, `context`, `gen`, `init`, and `sync-agent-guidance`, but nothing for `info`.

## What to do

Add `tests/test_cli_info.py` covering:

1. `--available` lists all four presets with their descriptions (`_show_available_tiers`).
2. In a `tmp_path` workspace with a valid `app.json`, the current-config path prints app name, preset, and the resolved feature flags.
3. `features` overrides in `app.json` win over the preset defaults (`info.py:98`).
4. Missing `app.json` prints the "Run 'mozaiks init <preset>'" hint.
5. Malformed JSON in `app.json` prints a readable error rather than raising.

Use `capsys` for output and `monkeypatch.chdir(tmp_path)` for the workspace. No MongoDB, Node, or LLM key required.

## Run

```bash
python -m pytest tests/test_cli_info.py -q --no-cov
```

See CONTRIBUTING.md for why `--no-cov` is the right flag for a narrow local slice.

PR #314, opened 2026-08-19
fix(windows): handle UnicodeEncodeError on legacy consoles

### LINKED ISSUE #307: bug(windows): CLI crashes with UnicodeEncodeError printing status glyphs on legacy consoles
## Problem

`mozaiks info` prints per-feature status with Unicode check/cross marks:

```
mozaiks_cli/commands/info.py:104
status = "OK" if enabled else "NO"   # actually uses the U+2713 / U+2717 glyphs
```

`mozaiks_cli/commands/gen.py` uses similar non-ASCII glyphs.

On Windows consoles still using a legacy code page (cp1252 / cp437 — common in older `cmd.exe` sessions and some CI runners), writing these raises `UnicodeEncodeError` and the command dies instead of printing its report. Mozaiks documents Windows/PowerShell as a first-class path, so this matters.

## Reproduce

```bash
PYTHONIOENCODING=cp1252 python -m mozaiks info
```

## Suggested fix

Add a small helper in `mozaiks_cli` that picks glyphs based on whether the active stdout encoding can represent them:

```python
def _supports_unicode() -> bool:
    enc = getattr(sys.stdout, "encoding", None) or "ascii"
    try:
        "✓✗".encode(enc)
    except (UnicodeEncodeError, LookupError):
        return False
    return True
```

Fall back to `[x]` / `[ ]` (or `+` / `-`). Use it in `info.py` and `gen.py` — grep for the non-ASCII glyphs to find all five call sites.

## Tests

Add a test that monkeypatches the stream encoding to `cp1252` and asserts the command completes and emits ASCII-only output.

Run with `python -m pytest tests/test_cli_info.py -q --no-cov`.

Youssef-AMARZOU:
## Summary

Fixed `mozaiks info` crashing with `UnicodeEncodeError` on Windows consoles using legacy code pages (cp1252/cp437).

## Changes

- Added `_supports_unicode()` helper to check if stdout can encode Unicode characters
- Fall back to `[x]`/`[ ]` on legacy Windows consoles

## Testing

```bash
PYTHONIOENCODING=cp1252 python -m mozaiks info
```

Fixes #307

### DIFF
--- a/mozaiks_cli/commands/info.py
+++ b/mozaiks_cli/commands/info.py
@@ -2,10 +2,21 @@
 mozaiks info - Show current configuration.
 """
 
+import sys
 import json
 from pathlib import Path
 
 from mozaiks_cli.workspace import resolve_active_app_root
+
+
+def _supports_unicode() -> bool:
+    """Check if stdout supports Unicode characters."""
+    enc = getattr(sys.stdout, "encoding", None) or "ascii"
+    try:
+        "✓✗".encode(enc)
+    except (UnicodeEncodeError, LookupError):
+        return False
+    return True
 
 # Tier definitions
 TIER_PRESETS = {
@@ -104,8 +115,12 @@ def _show_current_config():
     print(f"  Preset:        {preset}")
     print(f"  Auth Required: {auth_required}")
     print("\nEnabled Features:")
+    use_unicode = _supports_unicode()
     for feature, enabled in sorted(features.items()):
-        status = "✓" if enabled else "✗"
+        if use_unicode:
+            status = "✓" if enabled else "✗"
+        else:
+            status = "[x]" if enabled else "[ ]"
         print(f"    {status} {feature}")
 
     # Show upgrade path if not full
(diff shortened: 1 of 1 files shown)


elCaptnCode:
Thanks for this @Youssef-AMARZOU

The fix itself looks right: the encoding probe handles a missing `sys.stdout.encoding`, and catching `LookupError` alongside `UnicodeEncodeError` is the correct call for an unknown codec.

One thing will fail CI once it runs — the import block is unsorted, and `I` (isort) is in this repo's ruff select list:

```
I001 Import block is un-sorted or un-formatted
 --> mozaiks_cli/commands/info.py:5:1
```

`ruff check --fix mozaiks_cli/commands/info.py` sorts it (`import json` before `import sys`). That should be the only thing between this and a merge.

Two things I'd rather track separately than pile onto your first PR:

- #307 also mentions the glyphs in `gen.py`, which this doesn't touch.
- A regression test — monkeypatching the stream encoding to `cp1252` and asserting ASCII-only output. It fits naturally into `tests/test_cli_info.py`, which #309 covers.

Happy to merge on the ruff fix alone. If you'd like either follow-up, say so and I'll assign you; if not, no problem at all — they'll go back in the pool.

Issue #325, opened 2026-08-19
docs: CONTRIBUTING never explains how to run ruff or mypy locally


elCaptnCode:
## Problem

CI's `lint` job runs two tools that can fail a PR:

```yaml
- name: Run ruff
  run: ruff check . --output-format=github

- name: Run mypy
  run: mypy mozaiksai/ factory_app/ --ignore-missing-imports --disable-error-code=import-untyped
```

Neither is mentioned anywhere in `CONTRIBUTING.md`. Confirm:

```bash
grep -in "ruff\|mypy\|lint" CONTRIBUTING.md   # no results
```

The guide carefully explains how to run *tests* locally — including the `--no-cov` nuance — but a contributor who follows it exactly can still open a PR that fails CI on the very first lint step, with no idea which command to run to reproduce it.

## What to do

Add a **Linting and Type Checking** section to `CONTRIBUTING.md`, next to "Running Tests Locally", documenting the exact commands CI runs:

```bash
ruff check .
mypy mozaiksai/ factory_app/ --ignore-missing-imports --disable-error-code=import-untyped
```

Worth including:
- both come from the `dev` extra, so `pip install -e ".[dev]"` already provides them;
- `ruff check . --fix` auto-fixes much of what it reports;
- mypy's scope is `mozaiksai/` and `factory_app/` only — `mozaiks_cli/` and `tests/` are not type-checked in CI, which is useful to know before adding annotations nobody will verify.

Copy the flags verbatim from `.github/workflows/ci.yml` rather than from this issue, so the docs stay true to CI if the flags have changed by the time you pick this up.

## Verify

Run both commands on a clean checkout and confirm the documented invocations actually pass.

Docs-only. Related: the `ruff format` enforcement issue — coordinate the CONTRIBUTING edit if both are in flight.

PR #336, opened 2026-08-19
docs: correct CONTRIBUTING.md coverage gate to match CI (70%) (#317)

### LINKED ISSUE #317: docs: CONTRIBUTING says the coverage gate is 30%, CI enforces 70%
## Problem

`CONTRIBUTING.md` tells contributors the coverage gate is 30%:

> This repo's shared `pytest` configuration enforces a repository-wide minimum coverage threshold (`--cov-fail-under=30`, measured against `mozaiksai`).

That matches `pyproject.toml:130`. But CI overrides it — `.github/workflows/ci.yml` runs:

```yaml
pytest tests/ -v --cov=mozaiksai --cov-report=xml --cov-report=term-missing --cov-fail-under=70
```

So a contributor can pass locally at 30% and be failed by CI at 70%, with nothing in the docs explaining why.

## What to do

Pick one and make the three places agree:

- **Preferred:** document the real gate. Update the CONTRIBUTING paragraph to say the local default is 30% but **CI enforces 70%**, and that CI is authoritative (the guide already says this in the next paragraph — it just names the wrong number).
- Optionally raise `pyproject.toml` to 70 so local and CI match, if the suite actually clears 70 locally. Check before proposing this.

## Verify

```bash
grep -n "cov-fail-under" pyproject.toml .github/workflows/ci.yml CONTRIBUTING.md
```

All three should tell the same story when you are done. Docs-only change if you take the preferred option.

aditya226-sharma:
## Summary
Updates CONTRIBUTING.md to accurately reflect that CI enforces a 70% coverage gate, not 30%. Explains the discrepancy between the local pyproject.toml default (30%) and the CI override (70%), and clarifies that CI is authoritative.

## Changes
- Updated coverage gate note to mention both the local 30% default and the CI 70% requirement
- Added note that CI is authoritative with the specific 70% threshold
- Updated CI note to reference the 70% gate explicitly

## Why
Contributors could pass locally at 30% and be failed by CI at 70% with no explanation. This was confusing.

Closes #317

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -63,12 +63,15 @@ focused test file(s) that cover what you touched:
 python -m pytest tests/test_your_file.py -q
 ```
 
-**Coverage gate note:** This repo's shared `pytest` configuration enforces a
-repository-wide minimum coverage threshold (`--cov-fail-under=30`, measured
-against `mozaiksai`). Running a narrow test file against that global
-threshold will often report a coverage failure even when every test you ran
-passes. That failure reflects the coverage math for the whole package, not a
-problem with your change.
+**Coverage gate note:** The local `pyproject.toml` configuration sets a
+repository-wide minimum coverage threshold of 30% (`--cov-fail-under=30`,
+measured against `mozaiksai`). However, **CI enforces a stricter 70% gate**
+(`--cov-fail-under=70` in `.github/workflows/ci.yml`). CI is authoritative —
+your pull request must pass the 70% threshold regardless of local results.
+
+Running a narrow test file against the global threshold will often report a
+coverage failure even when every test you ran passes. That failure reflects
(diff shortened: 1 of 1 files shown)


elCaptnCode:
Thanks @aditya226-sharma, and apologies for the delay — CI hadn't started because GitHub holds workflow runs from first-time contributors until a maintainer approves them. Our side, not yours. Approved now.

The change is correct: `pyproject.toml` sets 30, `.github/workflows/ci.yml` overrides it to 70, and saying plainly which one is authoritative is the right fix. Reads well too.

Merging once the checks come back green.

One small thing for a follow-up if you're interested — `CONTRIBUTING.md` also doesn't document how to run `ruff` or `mypy` locally, even though CI enforces both and it's the first thing a PR tends to trip on. That's #325 if you'd like it.

PR #384, opened 2026-08-24
feat(adapters): declarative adapter registry with load-time conformance

elCaptnCode:
First vertical slice of the modular-adapter work: the **mechanism**, proven end to end and actually wired in. Ports and providers come next, on top of this.

## What it does

An app declares its adapters, and the loader verifies each one satisfies the port it claims **before startup completes**:

```yaml
# app/config/adapters.yaml
schema_version: mozaiks.adapters.v1
areas:
  dns:
    port: app.services.adapters.dns.port:DnsProviderPort
    active_env: DNS_ADAPTER
    providers:
      cloudflare: app.services.adapters.dns.cloudflare:CloudflareDnsAdapter
      stub: app.services.adapters.dns.stub:StubDnsAdapter
```

Exposed as `AppLoadResult.adapter_registry` and `app.state.adapter_registry`. **Apps with no `config/adapters.yaml` get `None` and are entirely unaffected — zero regression surface.**

## The design decision worth reviewing

**This module defines no ports and knows no providers.** Port *and* implementation are both named by import reference, so they can live in the app bundle, a capability pack, or a hosted product.

That is a deliberate correction. `mozaiksai/core/ports/ssl_provider.py` currently has **zero OSS consumers** — not exported from `ports/__init__.py`, imported by hosted code only as `# noqa: F401` decoration, conformance never checked by anything. It is an orphaned published type that commits us to an API after 1.0 for no benefit. (`ports/artifact_store.py` is in the same state, and is additionally duplicated by `core/artifacts/store.py`, which is what every real consumer imports.)

The test for whether a port belongs in OSS is **does OSS runtime code call through it** — `EntitlementPort` is checked at module dispatch, `OrchestrationPort` drives execution, so both qualify. DNS/SSL ports have no OSS caller. This loader lets those ports live beside their consumers *without* giving up load-time verification, which is what made putting them in OSS tempting in the first place.

## Conformance, and why not `issubclass`

`runtime_checkable` protocols only support `issubclass` for method-only protocols, and the failure is an opaque `TypeError`. Member presence via `__protocol_attrs__` instead yields:

```
adapters.yaml area 'dns' provider 'partial': <class 'PartialAdapter'> does not
satisfy 'DnsProviderPort' — missing delete_record
```

**Every declared provider is verified, not just the active one** — a declaration that cannot be satisfied is a defect regardless of which provider is switched on today.

## Fail-closed, unlike subscriptions

`subscriptions.yaml` degrades to a warning and disables enforcement. A broken adapter declaration **raises and stops startup** instead. An unresolvable adapter will fail at first use anyway; degrading silently here would reproduce precisely the swallowed-import failure mode this mechanism exists to prevent.

## Tests — 25, failure-mode weighted

Partial implementation rejected naming the missing member; inactive-but-broken provider still rejected; unrelated object rejected; missing module/symbol/port fail closed naming the reference; unknown root and area keys rejected; wrong schema version rejected; `active` and `active_env` naming undeclared providers fail closed; unknown lookup lists what *is* registered. Plus three `AppLoader` integration tests proving the registry is genuinely reachable — including that a nonconforming adapter fails app startup, and that an app without the file is unaffected.

## Validation

- `ruff check .` clean
- Full suite: **13,559 passed**, 78 skipped
- The single failure in that run (`test_platform_startup_calls_migration_application_after_index_application`) reproduces identically on clean `origin/main` and is **not** caused by this change. Root cause found: that test depends on an untracked `.env`, so it fails in any git worktree and passes in the main checkout. Not a main regression — a test-environment contract gap, noted below.

## Follow-ups this enables (not in this PR)

1. Delete the orphaned `ports/ssl_provider.py`; reconcile the duplicate `ArtifactStore`.
2. A DNS pack shipping port + adapter templates + `required_integrations`.
3. A pack conformance test across `build_context/`.
4. `mozaiks pack new` scaffolder and an `add-adapter` skill — the "make it easy" layer, which is skill/CLI work rather than a workflow.
5. Declare `.env`-dependent tests' prerequisites, so worktree runs stop producing false failures.

## Impact
| | |
|---|---|
| Layer | Universal substrate (`mozaiksai/core/adapters`) + app loader + platform host |
| App workspace impact | New optional `app/config/adapters.yaml`; absent = unchanged behavior |
| Compatibility risk | None — additive, no existing app declares adapters |
| Docs | `docs/architecture/foundations/adapter-registry.md` |

### DIFF
--- a/docs/architecture/foundations/adapter-registry.md
+++ b/docs/architecture/foundations/adapter-registry.md
@@ -1,0 +1,131 @@
+# Adapter Registry
+
+Status: **implemented mechanism.** `mozaiksai/core/adapters/registry.py` and the
+`config/adapters.yaml` contract described here are live and loaded at app
+startup. Provider ports and implementations are not shipped by this mechanism —
+see [Ports do not live here](#ports-do-not-live-here).
+
+## What this is
+
+A declarative way for an app to say which provider adapters it has, and a
+loader that **verifies each one satisfies the port it claims** before the app
+finishes starting.
+
+```yaml
+# app/config/adapters.yaml
+schema_version: mozaiks.adapters.v1
+areas:
+  dns:
+    port: app.services.adapters.dns.port:DnsProviderPort
+    active_env: DNS_ADAPTER      # optional; overrides `active` per deployment
+    active: cloudflare           # optional default
+    providers:
+      cloudflare: app.services.adapters.dns.cloudflare:CloudflareDnsAdapter
+      stub: app.services.adapters.dns.stub:StubDnsAdapter
+```
+
+At startup `AppLoader` resolves every declared reference, checks conformance,
+and exposes the result as `AppLoadResult.adapter_registry` — surfaced on the
+platform host as `app.state.adapter_registry`. An app with no
+`config/adapters.yaml` gets `None` and is entirely unaffected.
+
+## Ports do not live here
+
+This module defines **no ports and knows no providers**. Both the port and the
+implementation are named by import reference, so they may live in the app
+bundle, in a capability pack, or in a hosted product. The runtime never needs
+to know which.
+
+That is deliberate, and it is a correction of a pattern that did not work. A
+port the OSS runtime does not call through does not belong in OSS: it becomes
+an orphaned published type that commits the project to an API with no consumer
+and no verification. The test for whether a port belongs in `mozaiksai/core/ports/`
+is simply **does OSS runtime code call through it** — `EntitlementPort` is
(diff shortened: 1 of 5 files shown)


elCaptnCode:
Closing this stale implementation PR rather than carrying an old, non-mergeable architecture branch indefinitely. The declarative adapter-registry concept may still be useful, but the runtime/application contracts have moved substantially since this baseline. If prioritized later, restack the mechanism from current main and re-evaluate it against the present app-loader, semantic compiler, and hosted/OSS boundary rules.

PR #421, opened 2026-08-28
feat(semantics): ADR 0007 Slice 3 offline projection adapters

elCaptnCode:
## Summary

Implements ADR 0007 Slice 3 as offline, fail-closed comparison infrastructure only.

An independent review of the original exact head `a0e46d25892b6fdc3b0f01ec0c932612b57078aa` found material defects. Commit `c016092719f4397f5c8b009d566d9104bbb84306` corrects them without widening Slice 3 authority.

- accepts current AppGenerator/AppSchemaOutput, DesignDocs, subscription, module/page/route, AgentGenerator workflow-bundle, deployment, recorded-AppBuildPlan, build-context, and AppContext ownership shapes
- parses AgentGenerator `files[]` and its `orchestrator.yaml` triggers instead of an invented top-level trigger shape
- consumes `AppContextVersion.ownership_boundaries`; ownership evidence never fabricates graph authority
- projects only facts supported by SemanticGraph v1 and records unrepresentable application semantics as typed gaps
- requires action, event, capability, scope, and ownership references to resolve rather than inventing closure nodes
- rejects duplicate identities, contradictory observations, unknown source roots, and obsolete synthetic envelopes
- distinguishes invalid taxonomy grammar from valid-but-unpinned namespace entries
- independently derives source facts before graph construction and represented facts from the graph, then requires exact equality
- classifies every canonical source leaf exactly once and keeps coverage/gap paths aligned

## Review corrections

The original implementation used stale or invented producer shapes, allowed graph construction to supply missing reference targets, coalesced duplicate semantic identities, used circular fact-equivalence proof, and treated application-semantic payloads as nonsemantic. It also had incomplete import-side-effect guards and could misclassify valid unpinned taxonomy references. These defects are corrected in the current head.

## Validation

- focused Slice 3 suite: 23 passed
- Slice 1-3 semantic contract suite: 274 passed
- affected producer/consumer suite: 504 passed
- golden/materialization suite: 300 passed, 8 skipped
- production-readiness gate (quick, frontend skipped): 750 passed, 4 skipped; source hygiene passed
- `ruff check .`
- `git diff --check` (clean; checkout line-ending warnings only)
- DCO sign-off present
- GitHub exact-head CI: lint, frontend, visual regression, packaging, dependency audit, infrastructure build, DCO, security scans, and shards 0/2/3 pass
- GitHub shard 1 remains blocked by the known PR #420 host/process leak on the initial run and one failed-job rerun: `test_workspace_app_intelligence_acceptance` receives the temporary root created by `test_serve_normalizes_uppercase_log_level`; 3,482 other tests in that shard pass

No live models were used.

## Boundaries

No production generator, renderer, runtime host, loader, workflow, Studio, control-plane, persistence, promotion, refinement, ArtifactRevision, RefinementPatch, CompilationPlan, or ADR 0006 execution authority imports or invokes this offline module. No deployment, network, persistence, or live-model operation is introduced.

Exact base: `359c3e06b157c897c5d9f7a2717e4d48b688bb53`
Original independently reviewed head: `a0e46d25892b6fdc3b0f01ec0c932612b57078aa`
Corrected head: `c016092719f4397f5c8b009d566d9104bbb84306`

Keep draft and unmerged. Because the independent reviewer authored a material correction, this head requires a short independent follow-up review after PR #420 is resolved/rebased as applicable.

---

## Independent Stage A review + corrective delta (`c0160927` → `69783c0c`)

Reviewed at exact head `c0160927` by an agent that did not author the original slice, then rebased onto `f6a8b34c` (PR #420's squash + #387) and corrected. The rebase was conflict-free and both Slice 3 files were byte-identical to the reviewed head before any corrective edit.

### Findings and disposition

| # | Finding | Severity | Disposition |
|---|---|---|---|
| 1 | Purity test order-dependent and vacuous | High | **Fixed** |
| 2 | Purity property fails on a cold module cache | High | **Fixed** |
| 3 | `build_context`/`workflows` advertised with no projector | Medium | **Fixed** |
| 4 | Projection narrower than the page-schema API-path contract | Medium | **Fixed** |
| 5 | Duplicate root aliases don't fail closed | Low | **Fixed** |
| 6 | Real contract fields labelled "producer drift or a typo" | Low–Med | **Fixed** |
| 7 | Entitlement-gate module-only coupling | Low | **Reproduced, pinned by test** |
| — | "Generated apps' `domain.*` events can never be projected" | — | **Refuted, not a defect** |

### 1 + 2 — cold-cache purity

`project_semantic_graph()` defaulted its registry to `default_taxonomy_registry()`, which lazily imports the runtime layout registry and transport event contract **at call time**; that pulls in the workflow manager, which reads the workflow catalog from disk. Measured: **116 modules** added by one call, including 20 × `core.workflow.*`, `runtime.app.*`, and `ag2`.

The registry is now a **required pinned Slice 1 input**. Slice 1 stays the only taxonomy authority — no second registry, no copied entries, no eager-import or pre-warm workaround, and `mozaiksai/core/taxonomy.py` is untouched.

The old proof was itself broken: it failed run alone and its `sys.modules` assertion was vacuous in the passing run, because 22 earlier tests in the file had warmed the cache (CI shards by file, so it never surfaced). It is replaced by a **cold-subprocess proof** importing only the projection seam and Slice 1 primitives, guarding filesystem reads/writes, network, cwd and env mutation, asserting no prohibited module import and no input mutation.

```
alone:                    1 passed
first, then whole file:  28 passed
whole file:              28 passed
fresh cold subprocess:    1 passed
```

### 3 — truthful supported-root scope

`build_context` and `workflows` sat in `_SUPPORTED_ROOTS` with no projector, so a source containing only one failed with *"source contains no representable semantic identity"* — the wrong cause. Both are real provenance in recorded sources (`workflows` is documented in-file as *recorded AppBuildPlan execution metadata; not an AppBuildPlan field*) that declare no graph-v1 identity. Every leaf is now an explicit typed `UNSUPPORTED` gap, and a provenance-only source fails with an accurate diagnostic naming the root. No nodes are invented.

### 4 — full page-schema API-path contract

`AppPageSchema` permits any `/api/...` path (`page_schema.py:34` `_API_PATH_RE`); only `/api/modules/{module}/{action}` names a declared action. Projection hard-failed on everything else, so the committed `factory_app/app/ui/pages/notifications.yaml` (`/api/notifications`) could not be projected. Valid non-module bindings are now typed gaps; invalid paths still fail closed; no action or module target is invented. The committed page is tested directly.

### 5–7

Duplicate canonical root aliases now fail closed (`CONTRADICTORY`) instead of using one envelope and reporting the other as ~97 gaps. The unclassified-field diagnostic no longer calls genuine current contract fields "producer drift or a typo". The entitlement-gate coupling was reproduced: it already returns a precise typed missing-reference naming the `entitlement_gate` path and never invents the capability — both the open case and the closure with `subscriptions` are now covered by test.

### Validation

Slice 1–3 semantic suites + production-readiness + hygiene **329 passed**; producer/direct-consumer (AppGenerator materialization, DesignDocs, subscription contract, page schema, AgentGenerator brownfield) **46 passed**; build-context and workflow producer suites **645 passed**; Slice 3 file **28 passed**. `ruff check .` clean; `mypy mozaiksai/core/semantics/` clean (9 files, `--python-version 3.12` locally for NumPy stubs); `git diff --check` clean; DCO signed on all commits; exactly three files changed.

No later-slice behavior, production authority, capability advertisement, ADR 0006 execution, entitlement-policy change, or live model is involved.

---

## CI diagnosis at `69783c0c` and rebase to `36eb35fb`

The first run at `69783c0c` failed only shard 1, on `tests/test_studio_host_smoke.py::test_studio_endpoints_work_without_auth_user_id` (`assert 500 == 200`) — not a Slice 3 test. Delta-debug bisect over the 133 preceding files isolated a single polluter: `tests/test_cli_serve_log_level.py`.

**Pre-existing on main, proven both ways**: the pair fails identically on `f6a8b34c`, and it is dormant there only because the two files sit in different CI shards. Adding the Slice 3 test file re-shards the suite (`test_studio_host_smoke.py` moves shard 0 → shard 1), co-locating the pair. Mechanism: the polluter guards `serve.run()`'s env writes with `monkeypatch.delenv(name, raising=False)` on absent keys — the PR #418 delenv-absent defect — so `MOZAIKS_APP_WORKSPACE_PATH` leaks pointing at its tmp bundle, and the Studio smoke test then resolves that workspace and 500s. Verified empirically by reproducing the failure with only that key leaked.

Fixed at the source in **PR #423** (one test file, #418 set-then-delete pattern, now also guarding `MOZAIKS_HOST`), merged as `4d880e36`. This PR is rebased onto that main; the two Slice 3 code files remain **byte-identical** to the reviewed corrective delta at `69783c0c`, and the polluter+victim pair plus the Slice 3 file pass together locally (40 passed).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,16 @@ This project follows a practical pre-1.0 changelog format:
 
 ### Added
 
+- **Offline semantic projection comparison (ADR 0007 Slice 3)**: deterministic,
+  input-immutable adapters can now project existing AppGenerator, DesignDocs,
+  subscription, module, workflow, deployment, recorded-artifact, build-context,
+  and ownership evidence into candidate `SemanticGraph` documents for offline
+  corpus comparison. A machine-readable coverage report classifies every source
+  leaf as projected, deliberately non-semantic, or deferred and returns typed
+  gaps for unsupported facts. Production generators, hosts, loaders, Studio,
+  workflows, control-plane code, capability advertisement, and runtime authority
+  remain unchanged.
+
 - **Semantic-compiler contract layer (ADR 0007 Slice 2)**: strict, immutable,
   content-digested contracts for `ApplicationManifest`
   (`mozaiks.app_manifest.v1`), `SemanticGraph` (`mozaiks.semantic_graph.v1`),
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,14 +31,14 @@ This project follows a practical pre-1.0 changelog format:
 ### Added
 
 - **Offline semantic projection comparison (ADR 0007 Slice 3)**: deterministic,
-  input-immutable adapters can now project existing AppGenerator, DesignDocs,
-  subscription, module, workflow, deployment, recorded-artifact, build-context,
-  and ownership evidence into candidate `SemanticGraph` documents for offline
-  corpus comparison. A machine-readable coverage report classifies every source
-  leaf as projected, deliberately non-semantic, or deferred and returns typed
-  gaps for unsupported facts. Production generators, hosts, loaders, Studio,
-  workflows, control-plane code, capability advertisement, and runtime authority
-  remain unchanged.
+  input-immutable adapters accept current AppGenerator, DesignDocs,
+  subscription, module, page/route, AgentGenerator bundle, deployment,
+  recorded-AppBuildPlan, build-context, and AppContext ownership shapes. They
+  project only graph-v1 facts and report every other source fact through typed,
+  machine-readable coverage and gaps; unresolved action, event, capability,
+  scope, and ownership references fail closed instead of creating graph facts.
+  Production generators, hosts, loaders, Studio, workflows, control-plane code,
+  capability advertisement, and runtime authority remain unchanged.
 
 - **Semantic-compiler contract layer (ADR 0007 Slice 2)**: strict, immutable,
   content-digested contracts for `ApplicationManifest`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,7 +42,10 @@ This project follows a practical pre-1.0 changelog format:
   they declare no graph-v1 identity and cannot by themselves produce a graph.
   Page bindings outside `/api/modules/{module}/{action}` are retained as typed
   gaps rather than invented action targets, so any valid AppPageSchema api path
-  projects. Projection requires an explicitly pinned Slice 1
+  projects. Surface realization kinds that graph v1 cannot retain are likewise
+  precise typed gaps rather than falsely reported as represented, and custom
+  route identities retain their current AppSchema producer paths. Projection
+  requires an explicitly pinned Slice 1
   `TaxonomyRegistry`, which keeps the call free of runtime, workflow, and
   workflow-catalog side effects. Production generators, hosts, loaders, Studio,
   workflows, control-plane code, capability advertisement, and runtime
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,12 +33,20 @@ This project follows a practical pre-1.0 changelog format:
 - **Offline semantic projection comparison (ADR 0007 Slice 3)**: deterministic,
   input-immutable adapters accept current AppGenerator, DesignDocs,
   subscription, module, page/route, AgentGenerator bundle, deployment,
-  recorded-AppBuildPlan, build-context, and AppContext ownership shapes. They
-  project only graph-v1 facts and report every other source fact through typed,
+  recorded-AppBuildPlan, and AppContext ownership shapes. They project only
+  graph-v1 facts and report every other source fact through typed,
   machine-readable coverage and gaps; unresolved action, event, capability,
-  scope, and ownership references fail closed instead of creating graph facts.
-  Production generators, hosts, loaders, Studio, workflows, control-plane code,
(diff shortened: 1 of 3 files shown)


elCaptnCode:
Independent exact-head review completed at `cd1e3fec9d383ad18abf109735245ca90d816b4e` against current main `5b08077c416a9d58ec319d2b3d4c0fc9a8718c51`.

No material defect found. The corrective delta accurately classifies current non-graph-v1 `surface_kind` values as typed unsupported gaps and preserves custom-route producer provenance without inventing route/page semantics. Offline purity, explicit taxonomy authority, independent fact extraction, exact leaf classification, fail-closed duplicate/contradiction handling, reference closure, and the no-production-authority boundary remain intact.

Local validation: 286 semantic tests; 41 direct producer/consumer tests; 72 build-context/workflow contract tests; production-readiness 750 passed, 4 skipped; Ruff clean; scoped semantics mypy clean; source hygiene clean; diff scope/check clean; all commits signed off. The repository-wide mypy configuration also reports the unchanged pre-existing `acp_coding_provider.py` `elicitation_policy` mismatch, outside this PR.

Exact-head GitHub state: 14/14 checks successful, DCO successful, cleanly mergeable, auto-merge disabled.

PR #434, opened 2026-08-29
fix: sanitize datetime/BSON in collections_first_docs_full context

laibapafaftab-debug:
## Boundary confirmation

- [ ] This change stays within the open-source `mozaiks` repository and does not introduce private hosted-product logic, credentials, or internal-only URLs.

## Governance check

- [ ] This is an ordinary change with no new public schema, public workflow/prompt family, provider mutation path, authority bypass, eval artifact, learned optimization, or cross-customer intelligence.
- [ ] If this adds or changes a public contract, the contract is classified and versioned.
- [ ] If this touches a one-way-door area from `OSS_PUBLICATION_POLICY.md`, a short ADR is included.
- [ ] If this touches permissions, secrets, provider execution, payments, deployment, DNS, or production operations, the authority and review path are explicit.


elCaptnCode:
Hi @laibapafaftab-debug — thank you for taking the time to investigate this. Your diagnosis makes sense against the version of the code your branch started from.

Since that fork point, `main` has moved forward substantially and now sanitizes context variables before they enter AG2 workflow `knobs`, including datetimes and other non-JSON-native values. Could you please:

1. sync/rebase your branch onto current `main`;
2. confirm whether you can still reproduce the serialization failure; and
3. if it still occurs, add a focused regression test showing the remaining path and update the PR description with the reproduction and tests run?

There are also two mechanical items we can help with afterward: Ruff currently flags the import block, and both commits need DCO sign-off. GitHub is holding the remaining CI because this is your first contribution—that delay is on our side, not yours.

Thanks again for contributing. Even if the original failure was addressed while your branch was in progress, confirming whether another path remains would still be genuinely useful.

elCaptnCode:
Hi @laibapafaftab-debug — thanks for syncing your branch with current `main`.

I reviewed the remaining diff. `main` now sanitizes context values centrally at the AG2 network boundary before they enter workflow `knobs`, and the existing regression coverage includes nested `datetime` values. Because of that, the local sanitizer in `collections_first_docs_full` appears to be redundant.

Could you please confirm whether the original serialization error still reproduces on your updated branch?

- If it no longer reproduces, we’ll close this PR as superseded by the changes that landed while you were working.
- If it still reproduces, please provide the reproduction steps or traceback and add a focused regression test demonstrating the path that bypasses the central sanitizer.

You can hold off on the Ruff and DCO cleanup until we confirm that a remaining bug exists.

Thanks again—the original report helped us verify that this boundary is now covered.

PR #469, opened 2026-09-02
fix(runtime): normalize generated Mongo response identifiers

elCaptnCode:
## Defect (reproduced by the read-only product baseline)

Generated app modules return raw Mongo documents from `ctx.persistence`-backed repos. Generated services set a string `id` and rely on the driver to autogenerate the `ObjectId` `_id` (the canonical-example and capability-pack repo pattern), so every list/read action over such records handed an `ObjectId` to FastAPI's serializer — which has no BSON branch — producing a bare HTTP 500 with no error code. The same unnormalized `result.data` flows to profile panels, profile tabs, page hydration, and relationship rows. The executor's response-size gate used `json.dumps(default=str)`, which masked the problem instead of surfacing it.

## Correction — one closed transport contract at one boundary

Normalization happens at `ModuleExecutor`'s success boundary — the single choke point every module action result crosses regardless of transport or how the module obtained its data — via one canonical normalizer, `mozaiksai/core/runtime/composition/bson_safe.json_safe_bson`. The contract is **closed**: this is not a universal serializer, and values outside the domain fail with a typed error rather than being coerced.

**Values:**

- BSON `ObjectId` **values** → their stable 24-character hexadecimal strings wherever they occur *as values*: `_id`, ordinary fields, nested documents, and arrays;
- `Decimal128` → its lossless decimal string; BSON `Int64` → exact builtin `int`;
- datetime/date/time → ISO strings, `UUID` → string, finite `Decimal` → the FastAPI-equivalent encoding, `Enum` → its recursively normalized value, valid-UTF-8 `bytes` → text, pydantic models → their explicit JSON-mode dump;
- non-finite floats and Decimals (NaN/Infinity), malformed UTF-8, scalar subclasses, and every unsupported type fail closed with `MODULE_RESULT_NOT_JSON_SAFE` — no `str`/`repr` fallback, no payload contents in error messages.

**Keys:**

- JSON object keys must already be **exact strings**;
- an `ObjectId` key — like every other non-string mapping key — is rejected with `MODULE_RESULT_NOT_JSON_SAFE`;
- no key normalization or key coercion occurs, so no post-normalization key collision is possible; the key domain is closed before iteration.

**Containers:**

- exactly `dict`/`list`/`tuple` are traversed (tuples become lists, matching JSON array semantics); the type check runs **before** any `.items()` call or iteration;
- sets/frozensets (no deterministic JSON form), container subclasses, custom Mappings, and generators are rejected without executing their code;
- cyclic containers and nesting beyond the depth guard are rejected; **input documents are never mutated**.

Normalization runs before output-schema validation and the size gate so both observe the JSON-safe shape. The size gate measures the exact response bytes: the same compact-separator, `ensure_ascii=False`, UTF-8 encoding Starlette's `JSONResponse` emits, verified byte-for-byte against real `TestClient` bodies. Narrower candidates were rejected deliberately: fixing only the HTTP module route would leave the four `platform.py` surfaces and WebSocket delivery broken; fixing the persistence wrapper would miss `literal_collection`/alias paths and `inserted_id`, and would strip BSON types modules may legitimately use internally.

## Proofs

- Normalizer attack matrix: ObjectId in every *value* position; ObjectId and other non-string keys rejected typed; Decimal128 losslessness; Decimal NaN/Infinity, hostile Mapping/set/dict/list/tuple subclasses (methods proven never executed), generators, malformed bytes, throwing pydantic serializers, Enum with unsupported values — all typed, never raw; cross-`PYTHONHASHSEED` proof that no set serialization path exists; no input mutation.
- Executor boundary: Mongo-shaped list/get results JSON-serialize with string ids through `ModuleExecutor.execute`; invalid results return `success=False` with `MODULE_RESULT_NOT_JSON_SAFE`; exact-limit passes and one-byte-over fails on real wire bytes.
- **Real Mongo over the packaged platform host** (`MOZAIKS_RUN_REAL_MONGO_TESTS=1`): create → list → read → update → delete over POST and GET dispatch, the profile-panel embedding surface hydrated from real documents, an invalid module result surfacing as the controlled JSON error envelope (not an unhandled serialization crash), and the server serving the next request normally.
- Product acceptance battery at this head: `test_generated_app_functional_acceptance.py` + `test_canonical_example_apps.py` + both boundary files green. Ruff, scoped mypy, `git diff --check` clean; full suite result in PR comments.

## Boundaries

No semantics/5D/AG2/hosted/evaluation surface touched; no serialization framework grown — one bounded, closed-domain normalizer at one boundary.

## OSS Change Impact

- Layer changed: runtime module-execution boundary (`mozaiksai/core/runtime/composition/`).
- Build workflow impact: none.
- Runtime/platform impact: module action results are JSON-safe on every response surface; behavior changes only for values that previously produced a 500 or that are now rejected typed instead of silently mis-serialized.
- App workspace impact: generated apps' read paths work without regeneration.
- Tests run: listed above + full suite (result in PR comment).
- Docs updated: CHANGELOG.
- Compatibility risk: none for previously-working responses; ObjectId-bearing responses change from 500 to strings; results carrying sets, non-string keys, or other out-of-domain values now fail typed instead of nondeterministically serializing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,16 +16,20 @@ This project follows a practical pre-1.0 changelog format:
 
 - **Generated Mongo documents no longer break HTTP responses**: module action
   results are normalized at the ModuleExecutor boundary so BSON `ObjectId`
-  values (including `_id`, nested documents, list items, and mapping keys)
-  become their stable hex strings and `Decimal128` becomes its lossless
-  decimal string. Generated apps that list or read Mongo-backed records —
-  whose documents carry a driver-generated `ObjectId` `_id` — previously
-  produced a bare HTTP 500 at serialization on every read path (module
-  routes, profile panels/tabs, page hydration). Datetimes and other
+  *values* become their stable 24-character hexadecimal strings wherever
+  they occur as values — `_id`, ordinary fields, nested documents, and
+  arrays — and `Decimal128` values become their lossless decimal strings.
+  JSON object keys must already be exact strings: an `ObjectId` key, like
+  every other non-string mapping key, is rejected with a typed
+  `MODULE_RESULT_NOT_JSON_SAFE` outcome — no key normalization or coercion
+  occurs, so no post-normalization key collision is possible; the key domain
+  is closed before iteration. Generated apps that list or read Mongo-backed
+  records — whose documents carry a driver-generated `ObjectId` `_id` —
+  previously produced a bare HTTP 500 at serialization on every read path
+  (module routes, profile panels/tabs, page hydration). Datetimes and other
   JSON-encodable values keep their existing wire semantics through explicit
-  closed conversions, unknown types and non-string mapping keys fail closed
-  with a typed `MODULE_RESULT_NOT_JSON_SAFE` outcome (no silent key
-  collisions, no repr leaks), cyclic or absurdly nested results are rejected
+  closed conversions, unknown value types fail closed with the same typed
+  outcome (no repr leaks), cyclic or absurdly nested results are rejected
   instead of exhausting the stack, and the response-size gate now measures
   the exact strictly-encoded UTF-8 bytes (no `default=str` masking). Input
   documents are not mutated. The container domain is closed to exact
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,8 +22,13 @@ This project follows a practical pre-1.0 changelog format:
   whose documents carry a driver-generated `ObjectId` `_id` — previously
   produced a bare HTTP 500 at serialization on every read path (module
   routes, profile panels/tabs, page hydration). Datetimes and other
-  JSON-encodable values are untouched, unknown types are never coerced
-  through a repr, and input documents are not mutated.
+  JSON-encodable values keep their existing wire semantics through explicit
+  closed conversions, unknown types and non-string mapping keys fail closed
+  with a typed `MODULE_RESULT_NOT_JSON_SAFE` outcome (no silent key
+  collisions, no repr leaks), cyclic or absurdly nested results are rejected
+  instead of exhausting the stack, and the response-size gate now measures
+  the exact strictly-encoded UTF-8 bytes (no `default=str` masking). Input
+  documents are not mutated.
 
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,7 +28,18 @@ This project follows a practical pre-1.0 changelog format:
   collisions, no repr leaks), cyclic or absurdly nested results are rejected
   instead of exhausting the stack, and the response-size gate now measures
   the exact strictly-encoded UTF-8 bytes (no `default=str` masking). Input
-  documents are not mutated.
+  documents are not mutated. The container domain is closed to exact
+  `dict`/`list`/`tuple`: sets and frozensets (no deterministic JSON form),
+  container subclasses, custom Mappings, and generators are rejected before
+  any iteration can run hostile code, and BSON `Int64` converts to the exact
+  builtin int. Non-finite `Decimal` values (NaN/Infinity), failing pydantic
+  serializers, malformed UTF-8 bytes, and unexpected conversion failures all
+  surface as the same typed error — never a raw exception — with hostile
+  payload contents kept out of error messages. The size gate's encoding is
+  byte-for-byte identical to what Starlette's `JSONResponse` emits (compact
+  separators, `ensure_ascii=False`, UTF-8), verified against real
+  `TestClient` response bodies, so a result at the limit passes and one byte
+  over fails.
 
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,20 @@ This project follows a practical pre-1.0 changelog format:
 
 ## Unreleased
 
+### Fixed
+
+- **Generated Mongo documents no longer break HTTP responses**: module action
+  results are normalized at the ModuleExecutor boundary so BSON `ObjectId`
+  values (including `_id`, nested documents, list items, and mapping keys)
+  become their stable hex strings and `Decimal128` becomes its lossless
+  decimal string. Generated apps that list or read Mongo-backed records —
+  whose documents carry a driver-generated `ObjectId` `_id` — previously
+  produced a bare HTTP 500 at serialization on every read path (module
+  routes, profile panels/tabs, page hydration). Datetimes and other
+  JSON-encodable values are untouched, unknown types are never coerced
+  through a repr, and input documents are not mutated.
+
+
 ### Changed
 
 - Upgraded the AG2 runtime to 1.0.3, including ACP 0.12.1 compatibility and corrected usage-event accounting coverage.
--- a/mozaiksai/core/runtime/composition/bson_safe.py
+++ b/mozaiksai/core/runtime/composition/bson_safe.py
@@ -1,49 +1,133 @@
-"""Canonical BSON identifier normalization for module action results.
+"""Closed JSON transport contract for module action results.
 
-Module actions routinely return raw Mongo documents; ``ObjectId`` and
-``Decimal128`` are not JSON-encodable and previously escaped the executor to
-the FastAPI serializer as a bare HTTP 500. This is the one normalization
-authority for that boundary: only BSON-specific identifier/decimal types are
-converted. Values the platform's existing encoder already owns (``datetime``,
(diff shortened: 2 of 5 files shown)


elCaptnCode:
Full local suite at head `fe92400b`: 14186 passed, 82 skipped, 0 failed (the real-Mongo CRUD normalization test skips without MOZAIKS_RUN_REAL_MONGO_TESTS=1 and was run explicitly against real Mongo: passed).

elCaptnCode:
## Review correction pushed — requesting Codex 2 exact-head re-review

**Reviewed-defective head:** `fe92400b` → **corrected head:** `87abb484` (correction commit `0741bff3`, mechanically rebased onto current main `d0467218`; zero conflicts).

**Finding 1 — key collisions, fixed by closing the key domain.** Mapping keys must already be strings (Mongo field names are strings); ObjectId/int/bool/None/tuple/bytes/custom keys fail closed with the typed error — no conversion, therefore no collision, no silent overwrite. Regressions: ObjectId-key + equivalent string key, `1` + `"1"`, `None` + `"null"`, tuple/bool/custom keys.

**Finding 2 — unsupported values, fixed with a closed explicit value domain.** None/bool/int/finite-float/str/ObjectId→hex/Decimal128→lossless-decimal-string/string-keyed mappings/lists/tuples, plus FastAPI-equivalent explicit conversions for what existing modules already returned (datetime/date/time→ISO, UUID→str, Decimal via `fastapi.encoders.decimal_encoder`, Enum→value, pydantic→JSON dump, UTF-8 bytes, sets→lists). Unknown types and non-finite floats fail closed; no str/repr/vars fallback (regression proves no repr leak).

**Finding 3 — cycles/depth.** Active-traversal identity detects cyclic mappings/sequences → typed error; bounded depth 100; shared non-cyclic children encode deterministically per occurrence. No raw RecursionError.

**Finding 4 — size gate fail-open removed.** Exact order now: normalize → output-schema validation on the transport shape → strict `json.dumps(ensure_ascii=False, allow_nan=False)` → UTF-8 byte length → size limit → return. `default=str` and the except→0 fail-open are gone; strict-serialization failure returns `success=False, error_code=MODULE_RESULT_NOT_JSON_SAFE` — a transport-contract outcome distinct from `EXECUTION_ERROR`, with no object contents in the client message.

**Real HTTP proof (`test_generated_app_http_bson_safety.py`, real local Mongo + packaged platform host):** boots `platform.app` over a generated-style bundle whose module uses `ctx.persistence` (driver-generated ObjectId `_id`), drives create→list→read→update→delete plus the GET dispatch surface — every response HTTP 200 valid JSON with 24-char string `_id`s, zero bare 500s.

**Corpus (no regeneration):** canonical apps (project-hub / research-ops / reporting-saas), functional acceptance, materialized-bundle production runtime — 116 passed with existing wire behavior preserved. Contract + executor suites: 78 passed. Full suite at head: **14240 passed / 83 skipped / 0 failed**; ruff, scoped mypy, `git diff --check` clean.

Scope: bounded ModuleResult transport boundary only — no universal serializer, no AG2 envelopes/digests/persistence-internal documents/semantics/B2A/hosted/evaluation surface touched. Draft retained, auto-merge disabled; I will not merge.

elCaptnCode:
@Codex 2 — correction round complete. Requesting exact-head re-review of `ef893af42d6acfed8a2ff64f17014c6166780f11` (previous reviewed-defective head: `87abb484f12a8cb92b89124f5c36700a5049c800`).

All five findings corrected, each with attack-matrix proof (`tests/test_module_result_bson_normalization.py`, 54 tests):

1. **Decimal NaN/Infinity** — rejected before the FastAPI `decimal_encoder` runs (`Decimal.is_finite` via unbound method); `NaN`, `Infinity`, `-Infinity`, `sNaN` all raise only `ModuleResultNormalizationError`.
2. **Pydantic serialization failures** — `model_dump(mode="json")` is wrapped: an unsupported `Any` field and a throwing `field_serializer` both surface as the typed error, with serializer message contents kept out of the error text.
3. **Hostile Mapping/set subclasses** — the container domain is closed to exact `dict`/`list`/`tuple`, with the `type()` check preceding any `.items()` call or iteration. Tests instrument hostile `dict`/`list`/`tuple`/`set` subclasses and a custom `Mapping` and assert their methods were **never executed**. Scalar subclasses (int/str/float) are likewise outside the exact domain; the one driver-produced exception is `bson.Int64` → exact `int`. Conversion types use unbound base-class methods (`ObjectId.__str__`, `datetime.isoformat`, `bytes.decode`, `UUID.__str__`) so subclass overrides cannot inject code.
4. **Sets** — rejected entirely (no sort-and-accept). A subprocess test across `PYTHONHASHSEED` 0/1/424242 proves the rejection is byte-identical under every seed, i.e. no set serialization path exists.
5. **Wire-size parity** — the executor's size gate now encodes with `ensure_ascii=False, allow_nan=False, indent=None, separators=(",", ":")` — byte-for-byte what `starlette.responses.JSONResponse.render()` emits — and the parity is proven against **real TestClient response bodies** including `{"text": "é"}`, emoji, and nested multi-byte payloads. A result at the exact limit passes; one byte over fails; an unsupported value fails as `MODULE_RESULT_NOT_JSON_SAFE`, never as a size outcome.

Exception translation (469-C): every conversion path raises only `ModuleResultNormalizationError`; the top-level boundary wraps unexpected ordinary exceptions by exception type name only (proven: hostile Enum `value` property raising with secret contents — contents absent from the message) and does not catch `KeyboardInterrupt`/`SystemExit`/`BaseException`. Cycles, depth, and shared-child behavior preserved (469-E).

Order preserved (469-G): result → normalize (typed) → output-schema warning → exact wire render → exact byte-size gate → success.

Real product proof (469-I, `tests/test_generated_app_http_bson_safety.py`): the packaged platform host over real Mongo now proves POST + GET module dispatch, the **profile-panel embedding surface** hydrated from real Mongo documents (`/api/me/profile-panels`), an invalid module result surfacing over HTTP as the platform's controlled JSON error envelope — distinguishable from Starlette's plain-text unhandled-crash body — and the server serving the next request normally. One test-design fix along the way: module dispatch calls now carry `?app_id=` explicitly, because dispatch resolves scope from principal/query (falling back to `default`) while profile hydration resolves it from the app manifest — the two surfaces read the same app-scoped collection only when the scope is passed.

Validation: full suite 14266 passed / 83 skipped / 0 failed; corpus battery (canonical example apps + generated-app functional acceptance + both boundary files) 89 passed; ruff clean; scoped mypy clean; `git diff --check` clean. Branch still based on `d0467218` (current main); no rebase needed.

elCaptnCode:
@Codex 2 — documentation-only correction complete. Requesting the narrow documentation delta review of `2da79035b2c036c53a5ab2f692e719ee595f72f6` (previous verified head: `ef893af42d6acfed8a2ff64f17014c6166780f11`).

**Scope**: exactly one repository file changed (`CHANGELOG.md`) plus the PR body metadata. Blob-identity proof against the head you verified:

- `mozaiksai/core/runtime/composition/bson_safe.py` — `75776a9e` (identical)
- `mozaiksai/core/runtime/composition/module_executor.py` — `aad39a48` (identical)
- `tests/test_module_result_bson_normalization.py` — `b1795e5f` (identical)
- `tests/test_generated_app_http_bson_safety.py` — `7a8fd8f0` (identical)

**Corrected canonical wording** (CHANGELOG + PR body): BSON `ObjectId` **values** become stable 24-character hexadecimal strings wherever they occur as values (`_id`, ordinary fields, nested documents, arrays); JSON object keys must already be exact strings; an `ObjectId` key — like every other non-string mapping key — is rejected with `MODULE_RESULT_NOT_JSON_SAFE`; no key normalization or coercion occurs, so no post-normalization key collision is possible (the key domain is closed before iteration); sets/frozensets and arbitrary container subclasses are rejected; `Decimal128` uses the accepted lossless representation; the size gate uses the same compact UTF-8 bytes as the platform response. The PR body also had two other stale pre-correction claims now fixed: "mappings/lists/tuples rebuilt" → exact `dict`/`list`/`tuple` only, and "every other JSON-encodable value passes through unchanged" → the closed conversion list with typed rejection for everything outside it (explicitly not a universal serializer).

**Wording audit**: diff-wide search finds no remaining "ObjectId mapping key is converted" claim; the `bson_safe.py` docstring already stated the correct key contract and is untouched.

**Validation**: focused normalizer + real-HTTP batteries 55 passed (confirmatory — runtime blobs unchanged); ruff clean; `git diff --check` clean; DCO signed. PR remains draft, auto-merge disabled, unmerged.

PR #474, opened 2026-09-03
fix(artifacts): establish exact run-bound build persistence

elCaptnCode:
## Replacement PR A — run/build persistence contract

First of two bounded PRs restacking #468 (superseded, will be closed once both replacements are published). This PR **defines and persists the authority**: the exact server-owned relation `workflow_run_id -> build_id` and the persisted records that prove it. It contains **no workflow bridge code and no lifecycle emitter wiring** — `build.started`/`build.completed`/`build.failed` cut over to this authority in replacement PR B (`cc/run-bound-build-lifecycle`), created only after this PR reviews clean and merges.

## A1. BuildRecord identity and index safety

- Canonical identity `(app_id, build_family, build_key, version_number)` with matching counter identity, materialized through the canonical runtime index verifier before the store becomes available — a failed migration leaves no client behind that skips verification.
- Retired indexes are identified by their **complete historical definitions**, verified everywhere as a preflight before any drop executes. A same-name index with any other definition fails the store closed; no name-only deletion, ever.
- `av_unique_current` (partial unique on CURRENT per app/family/key) makes at-most-one-CURRENT **storage-enforced**; concurrent independent-client publishes get a typed `BuildRecordCurrentConflictError` (idempotent same-target retries unaffected). Proven on real Mongo including raw-driver bypass writers (20 tests).

## A2. RunBuildBinding

- One closed, digest-sealed `RunBuildBinding` (`mozaiks.run_build_binding.v1`, `mozaiksai/core/artifacts/run_build_binding.py`): schema version, server scope, app_id, workflow name, immutable `workflow_run_id`, canonical `build_id`, exact BuildRecord identity (record id, family, key, version), `bundle_digest` when established, and a sealing `binding_digest`.
- Created at exactly one point: the terminal build tool, at the moment the current run establishes its build (BuildRecord creation). **Never** derived from chat ID, latest session build, latest journey build, latest CURRENT record, or caller strings — a pre-build run has no binding, and absence is the truthful state.
- The relation is persisted twice by design and with one authority: stamped immutably on the BuildRecord at creation, and projected as the sealed binding through the existing server-owned session seam. No second build store.

## A3. AppContext lineage

[...]

PR #472, opened 2026-09-03
fix(subscriptions): fail closed on invalid present contract

elCaptnCode:
## Defect (fail-open authorization)

`AppLoader` caught `SubscriptionsLoadError` from a **present but invalid** `config/subscriptions.yaml`, logged `SUBSCRIPTIONS_CONFIG_INVALID … entitlement enforcement disabled`, and continued with `subscriptions_config=None`. The platform then wired `NoOpEntitlementAdapter`, which grants **every** `entitlement_gate` unconditionally — a malformed contract silently disabled all paid-feature enforcement.

## Correction — one loader, fail closed, no schema change

- `AppLoader` now wraps `SubscriptionsLoadError` as `AppLoadError`: an app that declares a subscription contract must load it or not load at all; platform startup/readiness fails.
- **Absent file unchanged**: valid non-SaaS app, `subscriptions_config=None`, `NoOpEntitlementAdapter` permitted.
- **Both `mozaiks.subscriptions.v1` and `mozaiks.subscriptions.v2` remain accepted, unchanged.** `load_subscriptions_config` stays the sole schema authority — no second parser, no v1/v2 conversion, no permissive fallback, no hosted overlay, no renderer code, no `ProductDef` change.

## Security regression (core proof)

`test_gated_action_never_exposed_under_disabled_enforcement`: an app declaring an `entitlement_gate`d action plus a malformed `subscriptions.yaml`. **Before** (verified against unmodified main by stashing the fix): the app loaded with `DID NOT RAISE` — enforcement disabled, gated action granted to everyone. **After**: `AppLoadError` — the action is never exposed under disabled enforcement. Control case: the same app with a valid config loads and the gate denies without a grant.

## App Zero compatibility

Read-only check (no mozaiks-app edits): the hosted product's live `app/config/subscriptions.yaml` (`mozaiks.subscriptions.v2`, products platform/ai/hosting/marketplace) loads unchanged through this loader. A **generic** multi-product v2 fixture — multi-product, `default_product_id`, per-product assignment stores/plans, root wallets/top-ups/pricing groups, no proprietary plan data — is part of the committed matrix.

## Test matrix

[...]

PR #476, opened 2026-09-04
fix(runtime): close server-owned session and bundle identity boundaries

elCaptnCode:
## Independent boundary hardening (valid before and after 5D)

Small hardening extracted from the frozen #474 review round. Both fixes are independent of publication sequencing and remain valid before and after the 5D ArtifactRevision/ApplicationPublication cutover. This PR contains **no** CURRENT bundle/AppContext publication redesign, no transaction/CAS publication, no run/build lifecycle emitters, and no ArtifactRevision/ApplicationPublication production wiring.

## A. Server-owned session field protection

`SERVER_OWNED_SESSION_FIELDS = {workflow_run_id, run_build_binding, build_terminal_receipt}` — the chat-session lifecycle authority fields — now have exactly one durable write path: the new privileged `persist_server_owned_session_fields` API, which is **lease-fenced** with `assert_chat_mutable` so a worker whose chat execution lease was lost to a successor cannot replace the successor run's lifecycle authority (this closes the fencing gap Codex 1 flagged on the frozen #474 seam at its introduction point).

Every generic session write path was audited and closed:

| Path | Behavior |
|---|---|
| `create_chat_session(extra_fields=...)` | **Rejects** reserved keys deterministically, and the validation runs **before** the idempotent existing-session return: a prohibited request never silently succeeds, whether the session does not exist, already exists, or races into existence — logged security diagnostic + `ValueError`, zero mutation (existing documents stay byte-equivalent). |
| `persist_context_variables` | **Drops** reserved keys with a logged security diagnostic (model/tool-authored context must not crash the run, matching existing protected-key policy). |
| `fetch_chat_session_extra_context` (replay merge) | Preserves privileged values past workflow-declared context policy without ever accepting generic writes. |
| `mark_chat_completed`, `update_last_artifact`, pending-input request save/clear, tool-call state, general-chat upsert, ui-state backfill | Audited: fixed non-reserved field names only — no caller-controlled keys reach `$set`. |
| Raw factory session writers (`_record_chat_session_build_registry_id`, journey seeding in the platform build_lifecycle helpers) | Audited: fixed non-reserved field names (`build_registry_id`, journey fields) only. |
| `app_registry` module | Audited: reads chat sessions only, never writes them. |

No second session store; the privileged path writes the same session document through the same authority.

## B. Exact canonical bundle-entry identity

The writer's contract was independently confirmed: the archive is created as `{bundle_name}.zip` and recorded in the manifest at `{bundle_name}/{bundle_name}.zip`; `bundle_name` is derived through a closed character filter and persisted in `commit_metadata.metadata["bundle_name"]`; the generated-file mapper can never produce `application/zip`. No new "is_bundle" boolean was needed.

`resolve_canonical_bundle_entry(record)` (in `mozaiksai/core/artifacts/models.py`) enforces the exact contract:

1. the record itself is the canonical application-bundle record — `build_family == build_key == "app_bundle"` (another family/key is never accepted because its manifest looks bundle-shaped);
2. the persisted `bundle_name` satisfies the **closed shared grammar** (`validate_canonical_bundle_name`: ASCII letters/digits/hyphen/underscore only — structurally excludes path separators, dot segments, traversal, absolute/drive paths, control characters; invalid names fail closed, never normalized);
3. exactly one `application/zip` candidate exists;
4. its path equals **exactly** `canonical_bundle_archive_path(bundle_name)` = `{bundle_name}/{bundle_name}.zip` — no two-segment shape check, no basename inference, no suffix search, no digest membership search, no first-ZIP-wins, no normalization fallback;
5. its `sha256` is a lowercase 64-hex digest.

`GeneratedApp/other.zip`, `GeneratedApp/evil.zip`, `GeneratedApp/not-a-zip`, `GeneratedApp/.`, `GeneratedApp/..`, `GeneratedApp/subdir/file.zip`, `other/GeneratedApp.zip`, the correct digest under a wrong path, wrong-family/key records, and duplicate canonical-looking entries all raise typed `CanonicalBundleEntryError`.

**Writer/resolver equivalence**: the writer's name derivation is now ASCII-constrained to the same grammar, the manifest archive path is produced through the same shared `canonical_bundle_archive_path` helper, and the writer refuses an on-disk archive whose name disagrees with `{bundle_name}.zip`. A permanent regression resolves a real writer-produced record and asserts writer path == resolver requirement, so archive naming cannot drift without the contract changing. The AppGenerator terminal tool still fails the build closed at registration on any digest mismatch.

## Proofs

- Real Mongo (`tests/test_server_owned_session_fields_real_mongo.py`): create-with-injection rejected with no document created; benign extra_fields unaffected; generic context injection dropped and never stored; privileged write + replay round-trip; non-reserved key rejection; missing-session fail-closed; **lease-fence: a lost-lease worker's privileged write raises `ChatLeaseLostError` before touching Mongo**; reserved-set exactness.
- Bundle identity (`tests/test_canonical_bundle_entry.py`): exact valid case; wrong family/key/both; the full exact-path attack list (evil.zip, not-a-zip, dot segments, subdir archives, wrong first segment, case variants); zero/duplicate/malformed-digest candidates; same digest under noncanonical paths; the closed name-grammar matrix (slashes, backslashes, traversal, absolute/drive paths, control characters, spaces, dots, non-ASCII); shared-formula equivalence; plus the real-writer equivalence and archive-name fail-closed regressions in the greenfield suite.
- Session creation matrix (real Mongo): new+reserved rejected with no document; new benign+reserved rejected atomically; **existing session + reserved (only and mixed) rejected with the existing document byte-equivalent**; existing + benign keeps the idempotent early return; privileged setter with valid lease succeeds; lost lease rejects before mutation; dotted reserved-key variants cannot materialize.
- Regressions: Genesis persistence suite (greenfield registration + generate_and_download persistence), #472 subscription fail-closed suite, #469 transport/BSON suites, chat-lease unit + real-Mongo suites, source-hygiene gate — all green.
- Full `pytest -q --no-cov`: **14,299 passed, 0 failed**. Ruff clean. `git diff --check` clean. Scoped mypy blocked locally only by the known environmental numpy-stub artifact (CI is the arbiter).

## Boundaries

Untouched: BuildRecord CURRENT promotion sequencing, AppContextVersion promotion sequencing, publication CAS, ArtifactRevision production authority, lifecycle event emitters, semantic compiler, AG2, subscriptions, hosted, mobile, evaluation.

## OSS Change Impact

- Layer changed: runtime persistence (session write boundary) + artifacts model layer (canonical bundle entry) + AppGenerator terminal tool (registration-time verification).
- Build workflow impact: AppGenerator bundle registration now fails closed on an ambiguous manifest.
- Runtime/platform impact: server-owned session fields are closed to generic writes; privileged writes are lease-fenced.
- App workspace impact: none.
- Tests run: listed above + full suite.
- Docs updated: CHANGELOG.
- Compatibility risk: pre-production internal contracts only.

Reviewer: Codex 1.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,29 +12,28 @@ This project follows a practical pre-1.0 changelog format:
 
 ## Unreleased
 
-### Added
-
-- **Canonical CompilationPlan authority contract (ADR 0007 lane)**: a plan
-  self-digest proves body integrity, not truthful derivation — a caller can
-  mutate any derived plan fact, recompute the digests, and obtain a
-  structurally cold-valid plan. `CompilationPlanAuthorityInputs` is the one
-  strict immutable authority-input document (semantic graph, complete payload
-  closure, canonical layout-registry snapshot, scope selection,
-  structured-output configuration documents), and
-  `validate_compilation_plan_against_authority` re-derives the candidate
-  through the single canonical `derive_compilation_plan` implementation,
-  requires exact execution-authorizing equality, and returns the canonical
-  rederived plan. No proof object, token, or bearer capability exists:
-  possession of nothing establishes validity, serialized authority inputs are
-  never trusted without full rederivation, and brownfield plans without a
-  base-input authority are rejected fail-closed. CompilationPlan identity
-  now pins `assignment_contracts_digest` — the exact assignment-descriptor
-  closure consulted during derivation, so only descriptors that can affect a
-  plan affect its identity; structured-output enum names are content-derived
-  (sha256) instead of process-salted `hash()`, making config-bearing plan
-  digests deterministic across processes and PYTHONHASHSEED values; and the
-  canonical derivation path no longer reads ambient assignment-registry
-  state. Consumer wiring is a separate enforcement change.
+### Fixed
+- **Server-owned session fields and the canonical bundle entry are closed
+  boundaries**: the chat-session lifecycle authority fields
+  (`workflow_run_id`, `run_build_binding`, `build_terminal_receipt`) are now
+  writable only through a privileged, lease-fenced persistence API — generic
+  session creation with `extra_fields` rejects them deterministically before
+  the idempotent existing-session return (a prohibited request never
+  silently succeeds, whether the session exists or not), generic context
+  persistence drops them with a logged security diagnostic, and the replay
+  merge preserves them past workflow-declared context policy without ever
+  accepting generic writes. A worker that lost its chat execution lease to a
+  successor can no longer replace the successor's lifecycle authority. The
+  app-bundle manifest now has an exact canonical bundle-entry identity: the
+  record must be the canonical `app_bundle`/`app_bundle` BuildRecord, its
+  persisted bundle name must satisfy the closed shared name grammar (ASCII
+  letters/digits/hyphen/underscore), and the single `application/zip` entry
+  must sit at exactly `{bundle_name}/{bundle_name}.zip` — one shared formula
+  used by both the manifest writer and the resolver. A digest may never be
+  validated against "any manifest entry with this digest"; wrong-family
+  records, wrong archive basenames, dot-segment names, duplicate archives,
+  digest-less entries, and mislocated archives all fail the build closed at
+  registration.
 
 ### Fixed
 - **A present but invalid `config/subscriptions.yaml` now fails application
--- a/factory_app/workflows/AppGenerator/tools/generate_and_download.py
+++ b/factory_app/workflows/AppGenerator/tools/generate_and_download.py
@@ -691,9 +691,22 @@ async def _register_app_bundle_artifact_version(
         if app_dir is not None and written_paths
         else []
     )
+    # The archive identity comes from the one shared canonical formula: the
+    # manifest records the bundle archive at exactly
+    # {bundle_name}/{bundle_name}.zip, and a zip whose on-disk name disagrees
+    # with the canonical archive name fails the build closed instead of
+    # persisting a manifest that cannot be resolved.
+    from mozaiksai.core.artifacts.models import canonical_bundle_archive_path
+
+    canonical_archive_path = canonical_bundle_archive_path(bundle_name)
+    if zip_path.name != f"{bundle_name}.zip":
+        raise RuntimeError(
+            f"bundle archive {zip_path.name!r} does not match the canonical "
+            f"archive name {bundle_name}.zip"
+        )
     files_manifest = [
         {
-            "path": f"{bundle_name}/{zip_path.name}",
+            "path": canonical_archive_path,
             "sha256": sha,
             "size_bytes": zip_path.stat().st_size,
             "content_type": "application/zip",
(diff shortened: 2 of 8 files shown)


elCaptnCode:
@Codex 1 — review requested at exact head `ce72f5da8d175c94b4dc5f22f7ea854bf0b2727c` (base: main `553b40db`).

This is the small independent hardening PR extracted from your #474 review round — only the two fixes that remain valid before and after the 5D atomic-publication cutover:

**A — server-owned session boundary**: `SERVER_OWNED_SESSION_FIELDS` (`workflow_run_id`, `run_build_binding`, `build_terminal_receipt`) writable only through the new privileged `persist_server_owned_session_fields`, which is **lease-fenced with `assert_chat_mutable`** (your P1 on the frozen #474 privileged-setter seam, addressed at its introduction point). `create_chat_session(extra_fields=...)` now rejects reserved keys outright with no document created; `persist_context_variables` drops them with a logged security diagnostic; the replay merge preserves privileged values without accepting generic writes. Full audit of the remaining session write paths (fixed-field writers, raw factory journey writers, app_registry reads) is in the PR body.

**B — exact canonical bundle identity**: `resolve_canonical_bundle_entry` in `mozaiksai/core/artifacts/models.py` — exactly one `application/zip` entry at the exact persisted `{bundle_name}/{archive}` path with a valid sha256, typed `CanonicalBundleEntryError` otherwise. Digest-equality against arbitrary manifest entries is now structurally impossible as bundle authority; the AppGenerator terminal tool fails the build closed at registration if the persisted record does not resolve. Closure adoption (`receipt.bundle_digest == canonical entry sha256`) lands when the frozen #474 stack rebases.

**Deliberately absent** (per the freeze): atomic bundle/AppContext publication, CAS, lifecycle emitters, ArtifactRevision wiring — the divergence you found is 5D scope and a read-only implementation preflight for it has been prepared separately.

Validation: real-Mongo boundary matrix incl. the lease-fence proof; bundle substitution/duplicate/missing matrix; Genesis + #472 + #469 regressions; full suite 14,294 passed / 0 failed; ruff + hygiene clean.

elCaptnCode:
@Codex 1 — re-review requested at exact head `9901d593a8d0cf437f0869353e1086e0839494c2` (previous reviewed-defective head `ce72f5da`, base unchanged at main `553b40db`). One narrow correction round for your two findings:

**Exact app-bundle record identity**: `resolve_canonical_bundle_entry` now verifies the BuildRecord itself first — `build_family == build_key == "app_bundle"` — before looking at the manifest; wrong family, wrong key, and both-wrong are permanent regressions alongside the positive control.

**Exact archive path**: the writer was independently re-confirmed (archive created as `{bundle_name}.zip`, manifest path `{bundle_name}/{bundle_name}.zip`). One shared formula, `canonical_bundle_archive_path(bundle_name)`, is now used by both the writer's manifest construction and the resolver's exact-equality requirement — no two-segment/suffix/basename/digest-membership heuristics remain. Your attack list (`evil.zip`, `not-a-zip`, `GeneratedApp/.`, `GeneratedApp/..`, subdir archives, wrong first segment, correct digest under wrong path, duplicates) is all rejected in the permanent matrix. The bundle-name grammar is the closed shared validator (`validate_canonical_bundle_name`, ASCII letters/digits/hyphen/underscore) — the writer's derivation filter is now ASCII-constrained to the same set, invalid persisted names fail closed (never normalized), and the writer refuses an on-disk archive whose name disagrees with `{bundle_name}.zip`. A real-writer equivalence regression (writer-produced record resolves; writer path == resolver requirement) locks the two together.

**Deterministic create rejection**: reserved-field validation in `create_chat_session` now runs before the idempotent existing-session return — reserved input rejects deterministically with zero mutation whether the session does not exist, exists, or races into existence. Real-Mongo matrix: new+reserved (no document), new benign+reserved (atomic), existing+reserved only, existing benign+reserved (existing document byte-equivalent), existing benign-only (idempotent success preserved), privileged setter with valid lease, lost-lease rejection before mutation. Dotted reserved-key variants reconfirmed unable to materialize; the rest of the session boundary is untouched per your clean findings.

Validation: focused suites 174 passed (real Mongo included); full `pytest -q --no-cov` 14,299 passed / 0 failed; ruff, hygiene, `git diff --check` clean. No publication/5D, compiler, AG2, lifecycle-emitter, or #474 scope entered.

elCaptnCode:
@Codex 1 — head correction: the exact head for re-review is now `b09d60372fe1c0d380d4a8a697a42863684ea55e`. CI mypy (correctly) flagged an `Any | None` argument into the typed shared path formula in the new resolver code on `9901d593`; the fix is a two-line type-safety change in `resolve_canonical_bundle_entry` (validate the persisted name as `str` before passing it to `canonical_bundle_archive_path`) — no behavioral change, amended into the same correction commit. Everything in the previous comment stands; focused matrix re-verified and full suite re-run green on the amended head.

PR #484, opened 2026-09-06
feat(semantics): add closed contracts for action requests

elCaptnCode:
ActionPayload now requires one canonical recursive request contract, replacing the shallow request_fields authority. An inputless action explicitly declares a non-null closed empty object; omission and the retired field reject.

Base: `e2713d64205079bad8c7b3dbc44186f26998e652` (exact #483 main; post-main CI 34003537267 terminal green before branch creation).
Exact reviewed head: `2d840a932c20222861276407629f6ac5162d826b`.

## Contract and migration

- One closed immutable algebra: null, strict scalar, homogeneous array, closed object, and required/optional property. Value nullability is independent of property absence. Scalar enums reject duplicates, null, mixed types, bool/integer coercion, and non-finite numbers; names and enum values canonicalize deterministically.
- Profile `mozaiks.closed_contract_profile.v1`: depth 32, 1024 contract occurrences. Iterative checks bound depth/width and reject cycles before recursive validation. There are no refs or general unions. INTEGER enum values use signed 64-bit integers; NUMBER enum values use exact finite floats.
- Nested models revalidate, all collections store tuples, and unchecked copy/update paths reject. Identity uses the existing canonical serialization primitives.
- The existing CanonicalJsonValue implementation moves unchanged to canonical_json.py. Plan authority imports the same classes; all seven extracted definitions remain AST-identical. Exact JSON bytes and the complete 59,230-byte plan-authority serialization remain unchanged.
- The only semantic producer outside fixtures is the offline projector. Its explicit bounded schema importer consumes module action input_schema only when closure and every assertion are supported. Open/omitted closure, unsupported assertions, refs, unions, cycles, and excessive limits fail UNSUPPORTED. Surface-only action names without request authority fail MISSING. Synthetic fixtures declare their requests; the recorded DesignDocs fixture remains untouched and now proves the precise missing-authority gap.
- No production prompt, generator, or runtime consumes request_fields. Production dispatch continues consuming module.yaml actions[].input_schema. Future realization must initially require normalized equality with the semantic request contract; no compatibility proof is implemented here.

## Identity evidence

Only three existing goldens change, all classified EXPECTED_SEMANTIC_MIGRATION:

| Identity | Before | After |
|---|---|---|
| Corpus graph | 19d4fa7f8db76e05a3710b58f5909ddb807ccb7ce37facb51e8278de2a19f86b | 421b7359cbd42d0f94a1ac9fbb38387240dda7bbea322add2a72027d14cc4810 |
[...]

PR #488, opened 2026-09-07
feat(semantics): verify implementation artifacts by exact content

elCaptnCode:
## What

Content-resolved implementation artifact authority + exact static module handler
export validation — the accepted prerequisite so a future ImplementationBinding v2
can truthfully claim **"this exact implementation realizes this semantic
workflow/action."** A `ChildContractRef`/digest alone is not sufficient; this PR
makes implementation selection prove exact scope, canonical artifact family,
canonical `ArtifactAddress`, exact verified bytes, exact document schema version,
exact parsed public contract, exact module action export, and exact selected
handler source(s).

Base: main `f90281c9ff151dcc33e8744f1be325de3c4b5a42` (#489 + #491 merged).
Head: `9d5cfe41` — includes the Codex 4 correction round (below).

## Correction round (Codex 4 Astra findings — all five fixed)

1. **Split authority is content-resolved, never caller-asserted.** The
   caller-constructible `WorkspaceHandlerSplitAuthority` token is REMOVED. The
   public resolution APIs (`resolve_module_action_implementation`,
   `resolve_module_base_handler_source`) accept only the exact scope-bound
   **pack-contract selection**; `resolve_workspace_handler_split_authority`
   derives contract type, canonical contract id, the canonical
   `build_context/{contract_id}/contract.yaml` path (from the PARSED bytes,
   so a contract cannot be smuggled under another pack`'`s path), and this
   module`'`s manifest/leaf/base ownership — all from
   `get_verified_blob`-resolved bytes. The resolved authority (an internal
   frozen dataclass carrying the contract digest, scope, contract id, module
   instance, and both paths) is stored on
   `ResolvedModuleActionImplementation.split_authority` for future pinning.
   Signature-guard test proves no `split_authority` parameter exists on any
   public resolution API.
2. **EXPLICIT_HANDLER = true standalone one-source class: ZERO bases.**
   `class Handler(Base)` / `class Handler(Mixin, Base)` never certify
   explicit, even with an explicit override of the selected method.
3. **Canonical base import placement is proven.** The single
   `from .base_handler import <Base>` must be an unconditional DIRECT member
   of the module body appearing BEFORE the handler class definition.
   Conditional / try / TYPE_CHECKING / nested-function / after-class imports
   all reject (permanent tests for each).
4. **Closed scope-aware binding grammar.** The rebinding detector now covers
   the full Python 3.12 binding surface: def/class/import/assign/ann-assign/
   aug-assign/delete/for/with targets, global/nonlocal, `type` aliases
   (guarded — the node exists on 3.12+; the same attack is a parse failure
   on 3.11), **exception-handler names**, **match-pattern captures**
   (`case name`, `[*name]`, `{**name}`), and **walrus targets** including
   comprehension positions and another function`'`s default arguments.
   Nested function/class/lambda scopes bind their own locals only;
   comprehension iteration targets stay comprehension-local.
5. **Execution-scope-bound source selection.** Handler, base-handler, and
   pack-contract sources are `SelectedAccountedArtifact` values
   (`ExecutionAccessScopeRef` + digest-mandatory `AccountedArtifact`); the
   selection scope must equal BOTH the requesting scope and the module
   selection`'`s scope. Same bytes + same module id under another tenant or
   workspace scope fail closed — tested for leaf, base, and contract
   independently.

**Corrected split-override rule** (supersedes the earlier body): a canonical
split leaf that explicitly defines the selected method remains a two-source
**CANONICAL_BASE_HANDLER** certification — both source digests and the
pack-contract digest stay meaning-bearing; the proof records
`method_source: handler | base_handler`. `EXPLICIT_HANDLER` is reserved for
zero-base standalone classes. There is no third mode.

## Implementation identity

`ModuleActionExportProof.implementation_digest` (canonical JSON digest):
- EXPLICIT_HANDLER: mode, method_source, handler class, handler method,
  handler digest — no irrelevant base fields.
- CANONICAL_BASE_HANDLER: those plus base class, base digest, and
  `split_contract_content_digest` — changing only the leaf, only the base,
  or only the certification authority changes identity (proven: 4 distinct
  identities). Module/action identity stays outside the digest for
  ImplementationBinding v2 to pin separately.

## Real canonical pack acceptance

`tests/test_pack_implementation_certification.py` independently discovers and
pins the corpus: **10 workspace_handler_split modules, 63 actions**. Every
action certifies end to end — #484 request import, exact manifest, split
authority from the real pack contract bytes, both-source verified bytes,
`method_source = base_handler` (no checked-in leaf overrides exist), all
three digests in the proof — and a counting content store proves the
pack-contract blob is READ for every certification (a zero-contract-read
certification is now impossible). The commerce `module.yaml` capability
repair from the previous round is unchanged (Codex 4: CLEAN), as are the
app-scope base-handler layout selections (CLEAN; the module-relative
registry row asymmetry remains deliberately out of scope).

## Hostile matrix (all rejecting, permanently tested)

Round-1 matrix unchanged. Split-authority: absent/tampered/missing-blob
contract bytes; non-YAML bytes at the correct filename; wrong ownership
(leaf not workspace / base not templates); required_outputs omitting the
module manifest; unrelated pack contracts; wrong canonical path; wrong path
scope; non-pack contract types; base/contract selections without each other.
Cross-scope: leaf, base, and contract each under another workspace or
tenant. Grammar: mixins, multiple/transitive inheritance, aliased/wrong-
module/star/dynamic/conditional/try/TYPE_CHECKING/nested/late imports,
monkeypatch, setattr, `__getattr__`, class/method rebinding, base alias
references, exception-handler captures, match-pattern captures (star and
mapping-rest), walrus-in-default smuggling, type-alias rebinding.

## Pre-Astra hardening (head `82317525`)

Three nonblocking observations from the independent Fable review, hardened
before final review:

1. **Split-capable certifier made internal.** `certify_module_action_export`
   is now `_certify_module_action_export` and is removed from the public
   surface; `resolve_module_action_implementation` is the ONLY public
   authority-producing module-action API, and `prove_module_action_export`
   (zero-base EXPLICIT_HANDLER) is the only standalone helper. The
   fabricated-authority path — building a
   `ResolvedWorkspaceHandlerSplitAuthority` in memory and obtaining a split
   proof with zero pack-contract blob reads — no longer exists on the public
   surface; a permanent test proves no exported function accepts that type.
2. **Direct dynamic-execution escapes closed.** In construction scope
   (module/class bodies, plus enclosing-scope expression positions of defs;
   class bodies are entered as import-time execution, function/lambda bodies
   stay out of scope by design), certification now rejects: any reference to
   a dynamic export primitive (`exec`/`eval`/`setattr`/`delattr`/`getattr`/
   `globals`/`vars`/`__import__`/`type`), `from builtins import` of any such
   primitive (or `*`), and any use of a name bound to the `builtins` module
   (`builtins.exec`, `import builtins as b; b.exec`, `__builtins__`, and
   `e = exec` rebinds). Hostile tests cover all seven reviewer attack forms
   against the handler class plus base-class/method equivalents. Documented
   truthfully as a bounded own-source export/binding proof — not a proof of
   arbitrary imported dependency behavior.
3. **Ambiguous/implicit `required_outputs` ownership rejected.** At the
   certification authority boundary the verified contract bytes must be
   unambiguous: non-mapping entries and ANY duplicate `required_outputs.path`
   reject (duplicate ownership declarations are structurally ambiguous), and
   authority-relevant ownership must be EXPLICIT — omitted/null/empty owners
   never default; the module manifest owner must be the exact canonical
   `templates`. Real-contract census first: no pack contract has duplicates
   and every relevant entry declares an explicit owner, so all 10 modules /
   63 actions still certify unchanged.

## Final finite correction (head `955fd312`)

Codex 4 Astra found one remaining effective-export escape (architecture
disposition A — the two-mode design stands): locally-defined callable bodies
are uninspected, yet a module-scope call (`mutate()`), local class
construction (`Mutator()`), or decorator application (`@decorate`,
`@(lambda c: ...)`) executes them DURING import and could rebind the
certified export.

Closed with a **finite construction grammar — no call graph, no dataflow, no
import following**: uninspected locally-defined callables (functions,
classes, bound lambdas) may exist, but any construction-time `Load`
reference to one fails closed — deliberately subsuming direct calls, alias
chains (`alias = mutate; alias()`), decorator application and factories,
local class construction, and argument laundering. Direct lambda invocation
and lambda decorators reject. Class bodies are scanned recursively against
their own locals plus every enclosing scope (sibling classes included);
`locals` joins the dynamic-primitive set. **Deferred runtime helpers remain
allowed**: a helper called only from a method body, an uninvoked local
function, a stored-but-never-construction-invoked lambda, and imported
external callables all stay certifiable (permanent positive tests). The
effective-export claim is now stated precisely in the ADR: within the
bounded own-source construction grammar, the certified visible binding IS
the import-time export; imported dependency behavior and deferred bodies
remain outside the proof. The 10/63 real corpus recertifies unchanged.

## Anonymous-lambda closure (head `c8c46a3c`)

Independent micro-review confirmed the named-local-callable rule sound but
found one remaining category: **anonymous lambda laundering** — wrapping a
lambda before invoking it (`(f := lambda: ...)()`, `[lambda: ...][0]()`,
`{0: lambda: ...}[0]()`, conditional expressions) evaded the direct
`Call.func is Lambda` check and could execute an uninspected body at import.

Closed with ONE finite syntactic whitelist, not per-case enumeration: a
construction-scope lambda is permitted only as the entire direct value of a
simple named binding (`helper = lambda: ...`, one plain `Name` target;
annotated form included). Any other construction-time lambda position —
walrus, call func/args/kwargs, containers, subscripts, conditionals,
decorators and factories, defaults, annotations, class bases,
comprehensions — rejects generically. Walrus-to-lambda is out of the
callable vocabulary entirely and rejects regardless of invocation. A stored
lambda becomes an ordinary local callable name owned by the Load-reference
rule (`helper()` at construction still rejects); lambdas inside deferred
method/function bodies remain runtime behavior outside the proof.

Permanent tests: the reviewer`'`s weaponized `globals().__setitem__`/`type()`
regression, the ten-form laundering matrix in EXPLICIT mode, representative
attacks in split `handler.py` and `base_handler.py`, and positives for
annotated bindings, stored-never-invoked lambdas, and deferred-body lambdas.
The 10/63 corpus recertifies unchanged — no canonical source uses
construction-time anonymous lambdas.

## Validation

- Both #488 suites: **163 tests** (public-surface guard, dynamic-execution
  hostiles, and ambiguous-contract hostiles included); real corpus 10/63
  green with proven pack-contract blob reads.
- Preservation slice (#484/#485/#486/#489/#491, layout registry, compilation
  plan): **583 passed**.
- ruff, scoped mypy (new module), strict MkDocs, governance guardrails,
  `git diff --check`: all clean. Exact-head CI on `82317525` is mandatory.

## Notes for reviewers

Start at `implementation_artifacts.py`: `SelectedAccountedArtifact`,
`resolve_workspace_handler_split_authority` (the new authority boundary),
`_statement_binds_name` + `_nested_binding_names` (the closed binding
grammar), `_prove_canonical_leaf_subclass` (import placement + override),
and `certify_module_action_export` (method_source disposition). The
pack-contract selection is a bounded workspace-root build-context INPUT
(no layout-registry row was added; the registry has no contract.yaml family
and adding one would ripple compilation-plan identity fixtures — the
resolver instead proves scope + derived canonical path + verified bytes).

Draft. DCO signed. Auto-merge intentionally NOT enabled — do not merge.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -88,9 +88,18 @@ This project follows a practical pre-1.0 changelog format:
   analysis is closed over the full Python binding grammar including
   exception-handler and match-pattern captures. The entire real canonical
   split-pack corpus — 10 modules, 63 actions — certifies end to end with
-  proven pack-contract blob reads. The commerce module manifest's capability
-  entries were repaired to the canonical `ModuleCapability` contract (unique
-  capability ids with `kind`/`title`), which module loading requires.
+  proven pack-contract blob reads. The split-capable certifier is internal:
+  `resolve_module_action_implementation` is the only public
+  authority-producing API (no exported function accepts a preconstructed
+  split authority), direct dynamic-execution primitives are rejected in
+  construction scope even through `builtins` access, aliased builtins
+  imports, or simple rebinding (a bounded own-source proof — not a proof of
+  arbitrary imported dependency behavior), and pack-contract
+  `required_outputs` must be unambiguous: duplicate paths and omitted or
+  non-canonical authority-relevant owners reject. The commerce module
+  manifest's capability entries were repaired to the canonical
+  `ModuleCapability` contract (unique capability ids with `kind`/`title`),
+  which module loading requires.
 
 - **Closed semantic action requests**: `ActionPayload.request_contract` replaces
   shallow request fields with one immutable, bounded contract algebra for null,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,6 +65,16 @@ This project follows a practical pre-1.0 changelog format:
 
 ### Added
 
+- **Content-resolved implementation artifact authority**: implementation
+  selection for the future ImplementationBinding v2 now proves exact scope,
+  canonical artifact family/address, exact verified bytes, and exact document
+  schema versions for `orchestrator.yaml`, `structured_outputs.yaml`, and
+  `module.yaml` — resolved only through the immutable blob store, never a
+  filesystem path. Structured-output contract refs resolve only against the
+  selected workflow's exact configuration, and module handler sources carry a
+  bounded static export proof: the declared handler class and action
+  `handler_method` must be explicitly present in the selected verified source.
+
 - **Closed semantic action requests**: `ActionPayload.request_contract` replaces
   shallow request fields with one immutable, bounded contract algebra for null,
   scalars, homogeneous arrays, and closed objects. Offline module-schema imports
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -71,17 +71,25 @@ This project follows a practical pre-1.0 changelog format:
   schema versions for `orchestrator.yaml`, `structured_outputs.yaml`, and
   `module.yaml` — resolved only through the immutable blob store, never a
   filesystem path. Structured-output contract refs resolve only against the
-  selected workflow's exact configuration, and module handler sources carry a
-  bounded static export proof with exactly two certification modes and no
-  general source closure: `EXPLICIT_HANDLER` (the selected `handler.py`
-  explicitly defines the action `handler_method`) and
-  `CANONICAL_BASE_HANDLER` (the capability-pack contract proves the canonical
-  workspace_handler_split ownership and the regenerated `base_handler.py`
-  class explicitly defines the method behind the preserved leaf subclass).
-  Certified implementation identity covers both source digests, and the
-  entire real canonical split-pack corpus — 10 modules, 63 actions —
-  certifies end to end. The commerce module manifest's capability entries
-  were repaired to the canonical `ModuleCapability` contract (unique
+  selected workflow's exact configuration. Handler, base-handler, and
+  pack-contract sources are scope-bound selections (execution scope must
+  equal the requesting scope and the module selection's scope), and module
+  handler sources carry a bounded static export proof with exactly two
+  certification modes and no general source closure: `EXPLICIT_HANDLER` (a
+  true standalone class — zero bases — explicitly defining the action
+  `handler_method`) and `CANONICAL_BASE_HANDLER` (the two-source
+  workspace_handler_split closure, whose ownership authority is
+  content-resolved from the exact verified bytes of the module's owning
+  capability-pack `contract.yaml`, never caller-asserted; the canonical base
+  import must be an unconditional top-level statement before the class, and
+  a leaf override of the selected method stays a two-source certification
+  with a recorded `method_source`). Certified implementation identity covers
+  the leaf digest, base digest, and pack-contract digest; rebinding
+  analysis is closed over the full Python binding grammar including
+  exception-handler and match-pattern captures. The entire real canonical
+  split-pack corpus — 10 modules, 63 actions — certifies end to end with
+  proven pack-contract blob reads. The commerce module manifest's capability
+  entries were repaired to the canonical `ModuleCapability` contract (unique
   capability ids with `kind`/`title`), which module loading requires.
 
 - **Closed semantic action requests**: `ActionPayload.request_contract` replaces
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -72,8 +72,17 @@ This project follows a practical pre-1.0 changelog format:
   `module.yaml` — resolved only through the immutable blob store, never a
   filesystem path. Structured-output contract refs resolve only against the
   selected workflow's exact configuration, and module handler sources carry a
-  bounded static export proof: the declared handler class and action
-  `handler_method` must be explicitly present in the selected verified source.
+  bounded static export proof with exactly two certification modes and no
+  general source closure: `EXPLICIT_HANDLER` (the selected `handler.py`
+  explicitly defines the action `handler_method`) and
+  `CANONICAL_BASE_HANDLER` (the capability-pack contract proves the canonical
+  workspace_handler_split ownership and the regenerated `base_handler.py`
+  class explicitly defines the method behind the preserved leaf subclass).
+  Certified implementation identity covers both source digests, and the
+  entire real canonical split-pack corpus — 10 modules, 63 actions —
+  certifies end to end. The commerce module manifest's capability entries
+  were repaired to the canonical `ModuleCapability` contract (unique
+  capability ids with `kind`/`title`), which module loading requires.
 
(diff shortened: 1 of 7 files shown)


elCaptnCode:
Final exact-head Astra micro-review of the anonymous-lambda closure is CLEAN with no blocking findings.

Reviewed head: c8c46a3cd80b4d29a50e750450ae99ce58f97a77
Base/current main: f90281c9ff151dcc33e8744f1be325de3c4b5a42
Reviewed delta: 955fd3122756100b43a5f6d717d55855b3aad44d..c8c46a3cd80b4d29a50e750450ae99ce58f97a77

The whitelist exempts only the exact Lambda node stored as the entire direct RHS of a single-name Assign or a plain-name AnnAssign. Other construction-time lambda positions reject through generic AST traversal, regardless of wrapper. Defaults and enclosing construction expressions remain checked; nested lambdas do not inherit the exemption. Stored lambdas enter the existing construction-time Name Load restriction, while deferred use remains valid. Conservative compound-statement rejection is consistent with this bounded grammar.

The original reviewer-category regression and representative standalone, split-leaf and split-base cases reject statically. No generated source was executed. The correction remains a finite syntax rule, without a call graph, dataflow engine, import following, arbitrary body inspection, or transitive source closure. No remaining anonymous-lambda laundering defect was found under the declared grammar.

Validation:
- Exact-head CI: terminal SUCCESS, https://github.com/BlocUnited-LLC/mozaiks/actions/runs/34246415222
- Exact-head DCO: terminal SUCCESS, https://github.com/BlocUnited-LLC/mozaiks/actions/runs/34246415398
- Lambda-focused existing tests: 24 passed; positive-boundary selection: 10 passed (30 unique cases across those two selections).
- Independent parser-only matrix: 13 allowed and 46 rejected cases passed, plus exact-node identity checks.
- Corpus census and real-action tests: 11 passed, certifying all 63 actions across the same 10 independently established modules. Build-context/template tree unchanged; no template modifications required.
- git diff --check passed; review worktree clean. Full suite not rerun; broad regression coverage comes from exact-head CI.

Scope was limited to the anonymous-lambda correction; prior reviewed dispositions remain intact. Exact refs were rechecked before posting. No source edits, commits, or merge. PR remains draft with auto-merge disabled.

PR #487, opened 2026-09-06
fix(runtime): preserve declared tool routing through AG2 Network

elCaptnCode:
Deterministic workflow tools could not update routing state that explicitly allowed `deterministic_tool`: the factory bridge rejected the write as `context_bridge` before the Network adapter could publish it. Separately, source-scoped `tool_called` conditions were invisible to AG2's native packet builder, so a successfully executed tool did not dispatch the next agent.

## Related issue

N/A — two generic runtime defects reproduced by the compatibility gate.

## Current-main verification

- [x] Started from exact green main `451caaff5ac1d558de517eeb604b8c325d5e4331`, after merged #486 post-main CI succeeded.
- [x] Reproduced both failures against the unchanged main tree: an actual callable produced tool events but its context write was rejected; source-scoped tool routing stayed text-only, and the empty-text turn timed out.

## Summary

The trusted factory callable boundary now establishes invocation provenance bound to the exact bridge, policy instance, and workflow/app/chat run. One authority helper resolves the declared mechanism before the existing early permission check. Each pending set/delete retains its authorized writer for a second policy check before packet publication. Caller writer strings, hidden-context replacement, unbound invocations, other runs/policies, and expired child-task scopes cannot establish deterministic attribution. Raw packet sets/deletes retain ordinary bridge attribution. `ContextAuthorityPolicy` permissions are unchanged.

`SourceScopedToolCalled` is now a native `ToolCalled` subtype, which AG2 1.0.3 recognizes through its packet-builder `isinstance` check. It retains `FromSpeaker` evaluation and its registered serialization tag. AG2 continues to interpret actual tool events and construct/fold workflow packets.

The acceptance harness uses `factory.create_agents`, the real OpenAI SDK/AG2 response-processing path, actual callable tools and `ToolCallEvent`/`ToolResultEvent`, serialized/rehydrated graphs, and the actual Network runner. Only the SDK HTTP transport is fake. Correct-source handoff, context-driven dispatch, empty text, wrong source/tool, text-only and result-only negatives, native precedence, and HITL pause/resume are covered. No paid provider request was needed.

## Tests run

[...]

PR #485, opened 2026-09-06
fix(compiler): make structured output identity provider-neutral

elCaptnCode:
Canonical structured-output references currently hash provider-patched schemas, so OpenAI wire formatting can change compiler identity while hiding optionality. Schema validation also reports success after schema construction or evaluation errors.

This change separates unpatched acceptance models from the existing strict provider wrapper. `StructuredOutputContractRef` v2 pins `mozaiks.structured_output_acceptance_profile.v1` and its canonical schema digest. Exact object closure is applied at model construction, including nested models. Build and cold resolution require explicit model authority; assignment compilation reads configurations and identity bindings from the same immutable plan authority inputs.

Runtime Draft 7 validation now normalizes, checks, and evaluates schemas in order. Each failed stage returns a deterministic diagnostic. Module input/event boundaries reject failures; malformed output schemas return `INVALID_OUTPUT_SCHEMA`. Explicit `{}` remains universal, and permitted absence is handled separately. Existing output value-invalid warning behavior is preserved.

Provider and authority proof:

- Provider inlining, object closure, required-field preparation, and wrapper generation cannot alter canonical identity. Actual property, optionality, type, nested structure, enum, item, and exactness changes do.
- Canonical/provider cache order, cache clearing, declaration ordering and fresh-process hash seeds preserve references and digests. Equality-alias enum declarations retain their existing order-sensitive meaning.
- Both ref APIs require explicit configuration and exact model IDs. The canonical compiler supplies descriptor authority; assignment compilation, admission and artifact construction are cold authority boundaries. Runtime defaults never validate canonical references.
- Existing immutable authority inputs already pin source documents; no additional schema snapshot is needed.

Identity migration:

- Exact-base evidence records the old provider-influenced schemas, executable plans, assignment and artifact result. Only the structured-output-dependent unit and enclosing identity references migrate (`EXPECTED_STRUCTURED_OUTPUT_IDENTITY_MIGRATION`). Restoring those references/digests recovers the captured records exactly.
- Configuration/descriptor authority, generated bytes, all 61 existing corpus units, historical 59-unit proof, #483 workflow interface bytes/reuse, and #484 action request/canonical JSON identities remain unchanged. Existing fixed goldens are not repinned. ArtifactRevision evidence shapes are unchanged.

[...]

PR #491, opened 2026-09-07
fix(runtime): enforce exact structured-output auto-tool contracts

elCaptnCode:
## What

Fixes the two generic OSS runtime defects the independent App Zero Astra
runtime review exposed, without any App Zero workaround. Base: verified main
`5d51e740`. **CORRECTION ROUND 2 applied** — head `05130f16` supersedes the
round-1 reviewed head `c201479b` (which superseded `7d462758`). The two
remaining material findings are fixed; every area Astra confirmed clean is
preserved unchanged.

## Accepted architecture (confirmed clean, unchanged)

Declared structured outputs are exact at runtime (top-level/nested extras
reject before lossy normalization; open `dict`/`optional_dict` stay open
internally; provider wire models parse `extra="forbid"`); `structured_output`
is RESERVED runtime vocabulary (declaration rejection at both canonical
validators, no metadata override) served as a transient runtime-owned
read-only overlay (enumeration/snapshots never expose it); per-binding
payload isolation; context-only + explicit-param auto tools;
`include_auto_only`; real ThemeCapture Factory execution; #485 identity;
task batches; sequential/concurrent duplicate suppression.

## Round 2 — Defect A: callable authority is never cached

Declarative binding metadata may be cached; **Python callable authority may
not be cached across dispatches**. The binding cache now holds
`AutoToolBindingDescriptor`s only (agent, exact model identity, tool
declaration, ui config), self-validated against the live registry +
tools.yaml bytes. The executable function is resolved fresh through the
canonical workflow tool loader at EVERY dispatch (parameter metadata derived
from the fresh signature). Tool-file byte fingerprinting is removed — bytes
cannot prove callable behavior when an imported helper changes — and no
import/AST/transitive source-closure fingerprinting was added.

The canonical loader (`load_agent_tool_functions`) now refreshes
workflow-owned modules under every dotted package chain that reaches the
workflows root: the synthetic `workflows.<name>` namespace AND the real chain
derived from the filesystem (`__init__.py` walk — e.g.
`factory_app.workflows.<name>` plus the shared `_shared` helpers). Unrelated
workflows are not cleared. **External installed third-party packages are a
documented process-restart boundary** — never hot-replaced.

## Round 2 — Defect B: per-binding execution checkpoints

Each `(chat_id, turn_idempotency_key, binding)` tracks a finite process-local
checkpoint: PENDING → TOOL_TERMINAL → COMPLETE, with per-stage marks
(tool-call emission, context write-back, persistence, result emission). The
terminal record — detached result payload, status, and context snapshot — is
stored **synchronously after the tool returns, before the next cancellable
await**. Invariant: once a tool returns a truthful terminal result (success
OR failure), that binding is never invoked again for the same turn; retries
resume at the first unfinished binding/stage. Pre-execution interruption
still releases the claim so a retry may execute. Checkpoints pop on turn
completion and are bounded by `_CACHE_LIMIT` without evicting active
in-flight state. Documented explicitly as **process-local runtime
idempotency, not distributed exactly-once delivery** (crash-after-effect
before durable recording is the owning service`'`s idempotency contract).

## Round-2 proofs (tests/test_auto_tool_binding_cache_and_concurrency.py, now 22)

- Imported workflow helper changes observed on the next dispatch with
  byte-identical tools.yaml + tool file and NO reload — in the synthetic
  `workflows.*` namespace and the derived real package namespace; unrelated
  workflows` helper modules untouched; fresh-callable-per-dispatch with
  descriptor-cache reuse proven; tool-code change without reload observed.
- Cancellation attack matrix with deterministic barriers (no sleeps): cancel
  before invocation (retry executes, 1 total invocation), while blocked in
  the tool (pre-terminal, retry may execute), during persistence after tool
  success (**total tool invocations = 1**, persistence retried, result
  emitted once, tool-call not re-emitted), during result emission (no
  re-invoke, no re-persist, emission retried), multi-binding A/B/C with
  cancel after B`'`s result (A and B never re-run, B post-processing resumed,
  C exactly once), truthful tool FAILURE terminal (never re-invoked because
  it failed), bounded checkpoint cleanup that never evicts active state.

## Validation

Round-2 battery (auto-tool suites, loader/reload, cache invalidation, tool
routing #487, agent-factory web tools, context exposure, declarative
fallbacks, fail-closed, context authority + hostile + compile, task batches,
structured-output identity/provider/migration #485/#486, ThemeCapture,
capability packs #489): **583 passed**. Full `pytest -q --no-cov`:
**15,520 passed, 93 skipped**. ruff, scoped mypy (handler + loader), strict
MkDocs, governance guardrails, source hygiene, `git diff --check`: all clean.

## Docs

Auto-tool execution doc gains **Callable freshness** and **Execution
idempotency** sections (restart boundary + process-local crash boundary);
ADR 0007 and CHANGELOG updated to the round-2 invariants.

## Reviewer handoff (round 3)

Review head `05130f16`. Start at `_resolve_bindings` (fresh callable
resolution) + `_load_binding_descriptors` (declarative fingerprint), then
`_workflow_root_package_names`/`_reset_workflow_package_namespace` in
`agents/tools.py`, then the checkpoint machine in `_dispatch_claimed_turn`
(`_BindingExecutionRecord`). Highest-leverage questions: (1) is clearing
`<root>._shared` on every workflow load acceptably scoped for concurrent
in-process workflows sharing those helpers? (2) the tool-call emission stage
is marked complete only in the terminal record — is re-emitting on
pre-execution retry acceptable event-order truthfulness? (3) per-dispatch
loader cost (fresh module exec per auto-tool event) — acceptable for current
event rates?

Draft. DCO signed. Auto-merge remains disabled — do not merge.

### DIFF
--- a/.agents/skills/create-workflow/SKILL.md
+++ b/.agents/skills/create-workflow/SKILL.md
@@ -144,6 +144,11 @@ Rules:
 
 Declare the workflow state and which agents can see which variables.
 
+`structured_output` is RESERVED runtime vocabulary: declaring it (in
+`definitions` or agent `variables`) is rejected at workflow load. Auto tools
+receive it transiently via `context_variables.get("structured_output")` with
+no declaration.
+
 ```yaml
 definitions:
   artifact_request:
--- a/.claude/skills/create-workflow/SKILL.md
+++ b/.claude/skills/create-workflow/SKILL.md
@@ -142,6 +142,11 @@ Rules:
 
 Declare the workflow state and which agents can see which variables.
 
+`structured_output` is RESERVED runtime vocabulary: declaring it (in
+`definitions` or agent `variables`) is rejected at workflow load. Auto tools
+receive it transiently via `context_variables.get("structured_output")` with
+no declaration.
+
 ```yaml
 definitions:
   artifact_request:
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -280,7 +280,9 @@ products, and customer workspaces use the same shape:
   owning workflow's `ui/` folder.
 - `context_variables.yaml` declares runtime/session state. Large static prompt
   catalogs are injected by deterministic hooks; do not stuff them into context
-  variables.
+  variables. `structured_output` is reserved runtime vocabulary — auto tools
+  receive it as a transient read-only projection; declaring it as a context
+  variable fails workflow validation.
 - Reusable OSS build packs are named build contexts. `context.yaml` owns the
   pack descriptor through a `pack:` section plus capabilities and facades. It
   should be useful LLM/build context, not a placeholder
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,13 +29,23 @@ This project follows a practical pre-1.0 changelog format:
   `structured_output` is reserved runtime vocabulary: declaring it in
   `context_variables.yaml` now fails workflow validation with no metadata
   override, and overlay enumeration/snapshots never expose a colliding stale
-  base key. Auto-tool binding caches self-validate against the live
-  structured-output registry and tool declarations (reload, unload,
-  refresh_all, and failed reloads can never execute a stale binding or stale
-  callable), each binding receives its own detached copy of the validated
+  base key. Auto-tool binding caches hold declarative metadata only,
+  self-validated against the live structured-output registry and tool
+  declarations (reload, unload, refresh_all, and failed reloads can never
+  execute a stale binding); the executable tool callable is resolved fresh
+  through the canonical loader on every dispatch, so workflow-owned tool and
+  imported helper changes (in both the `workflows.*` and derived real package
+  namespaces such as `factory_app.workflows.*`) are observed without restart,
+  while external installed packages remain a documented process-restart
+  boundary. Each binding receives its own detached copy of the validated
   payload (explicit-argument mutation cannot contaminate another binding or
-  the audit record), and turns are claimed in-flight atomically so concurrent
-  duplicate deliveries produce exactly one tool side effect.
+  the audit record), turns are claimed in-flight atomically, and every
+  binding keeps a finite process-local execution checkpoint: once a tool
+  returns a truthful terminal result (success or failure) it is never
+  re-invoked for that turn — retries after cancellation resume unfinished
+  post-processing stages (write-back, persistence, result emission) instead
+  of re-running the tool. This is process-local runtime idempotency, not
+  distributed exactly-once delivery.
 
 ### Changed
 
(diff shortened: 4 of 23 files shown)


elCaptnCode:
**Astra Ultra — final exact-head review (round 2): CLEAN**

Reviewed head: `05130f164b02ef08a7fc4a9a7e6b46c1f2380463` (base `5d51e740`; delta from round-1 head `c201479b` reviewed in full). All 14 CI check-runs and DCO on the exact head verified terminal SUCCESS before review.

The final exact-head review found **no blocking issues**. Dispositions:

- **Callable freshness**: no executable callable cached across dispatches; descriptors are declarative-only; all four change cases (tool source, imported helper, tools.yaml, structured-output model) resolve current authority at the next dispatch without reload. PASS.
- **Namespace refresh**: synthetic `workflows.*` and filesystem-derived real chains (`factory_app.workflows.*` + `_shared`) both refreshed; unrelated workflows untouched; no stale-parent-attribute import form exists in the corpus; external packages remain a documented restart boundary. PASS.
- **`_shared` lazy-import skew**: not material — the documented contract resolves fresh workflow-owned code at dispatch boundaries; reset/import is single-loop synchronous, so no partially initialized package is observable.
- **Checkpoints**: TOOL_TERMINAL (success or truthful failure) is never re-invoked for the same turn/binding across cancellation, unexpected interruption, retry, and multi-binding execution (A/B/C = 1/1/1 with truthful ordering). Terminal records are detached in production paths; the `_detached_or_original` fallback is read-only downstream (nonblocking hardening note).
- **Write-back**: authoritative mutations happen synchronously through the overlay during tool execution (failures become truthful tool errors); the later write-back is a redundant detached re-copy — nothing can be lost irrecoverably. PASS.
- **Persistence unknown-completion window**: retry overwrites the same durable session record (`$set`, same identity/payload); only `session_version` staleness metadata increments again; no watcher, billing, entitlement, or transition triggers on session updates. PASS.
- **Result/tool-call emission unknown-completion window**: broadcast-only, key-stable (`call_id`/`tool_call_id` = turn key), no durable history on this path, no server-side state-mutating subscriber; UI suppresses successful auto-tool responses and resolves surfaces by id. Consumer-side dedupe noted as optional future hardening, not blocking. PASS.
- **Process-local boundary**: truthfully documented (no distributed exactly-once claim). Bounded checkpoint state verified (completed popped; in-flight never evicted; eviction horizon mirrors the accepted processed-set bound).
- **Performance watchpoint (nonblocking)**: fresh callable loading ≈4 ms (ThemeCapture), ≈100–145 ms warm / ~2 s first-load (AppGenerator) — acceptable at current auto-tool event rates.

Verification run at the exact head: 37 cache/concurrency + contract tests, 4 independent commit-before-return probes, and the bounded regression slice (nested-extra rejection, reserved-key rejection, real ThemeCapture path, task-batch smoke, #485 identity) — all green.

Not merged; merge authority remains with the maintainer.

PR #504, opened 2026-09-10
fix(runtime): preserve AG2 model output and retry limits

elCaptnCode:
## Summary
- Forward provider-native output token limits and SDK retry counts into the existing AG2 config adapter; selected provider entries override shared defaults.
- Reject unsupported, ambiguous, or malformed limits before client creation. Omitted limits retain AG2 defaults; zero retries is preserved.
- Verify real AG2 configs for OpenAI, Responses, Anthropic, Gemini, and Ollama, plus real SDK HTTP payloads and a single HTTP attempt on rate limiting with SDK retries disabled. Factory network acceptance also checks output limits.

## Scope and Safety
This is a necessary model-admission foundation, NOT a durable wallet reservation, USD cap, public free-tier enablement, provider retry replacement, or new billing subsystem. Current balance preflight and post-response debit cannot guarantee no overspend; docs state the remaining gap explicitly. No paid calls or infrastructure provisioned.

## OSS Change Impact
- Layer: runtime execution adapter (ExecutionEngine configuration).
- Build workflow: existing factory agents receive the declared controls; no graph, YAML contract, or prompt changes.
- App workspace: generic behavior inherited when upgrading the package; no hosted commercial rules.
- Compatibility: invalid/unsupported explicit limits now fail closed rather than being silently dropped. Native provider field names are not aliases.
- Docs: token management and Unreleased changelog.

## Tests
- Red regression proof: 34 failed / 2 passed before fix.
- Focused adapter/media/config/network: 77 passed.
- Full pytest -q --no-cov with isolated local Mongo: 16493 passed / 96 skipped / 4 warnings (827s).
- Ruff check . and git diff --check passed.
- Public release hold preserved.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,13 @@ This project follows a practical pre-1.0 changelog format:
 
 ## Unreleased
 
+### Fixed
+
+- Preserve explicit provider-native output token limits and SDK retry counts
+  when converting runtime LLM configuration to AG2. Invalid, conflicting, or
+  unsupported limits now fail before a provider request instead of disappearing.
+  This does not introduce a wallet reservation or a total-spend guarantee.
+
 ### App workspace alignment
 
 - Expose the existing reference bundle evaluator through `factory_app.eval` so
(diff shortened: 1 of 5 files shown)


elCaptnCode:
Follow-up e1915f7d keeps the same validated option values but passes them through the existing typed keyword-bundle pattern, resolving the CI mypy constructor errors. Focused 77-test slice rerun green. Full mypy mozaiksai/ factory_app/ now passes (604 source files) after installing the repo-declared types-PyYAML dev dependency locally. Ruff remains green. Full runtime regression result before this type-only fix: 16493 passed / 96 skipped.

PR #521, opened 2026-09-12
fix(factory): unblock live greenfield Genesis past the execution-identity guard

elCaptnCode:
## Why (DRAFT — do not merge)

Acting as a live builder, this branch ran real greenfield **Genesis** traversals against the Studio host with real OpenAI-backed AG2 agents. Each fix advanced the live run to the next stage. The structured-output/persistence blocker that previously kept this PR from producing any canonical artifact is now **closed and proven live**; the remaining scope (a full non-empty Revision 1, SecurityReadiness/AppReview over a real bundle, cross-build isolation, Revision 2) is **not yet proven** and is itemized below.

## 1. Target-identity defects (closed earlier on this branch)

1. **Execution-identity guard vs. registration hook (critical).** After #507 made execution identity immutable at the `before_chat` boundary, `ValueEngine.create_app_record` rebinding `app_id` → `draft-build-*` made **every greenfield build fail in ~1s** (`ContextAuthorityError lifecycle_changed_runtime_authority key=app_id`). Fix: the generated app's registry identity travels in a declared `generated_app_id` context var; the hook never rebinds the immutable executing `app_id`. Factory artifact/staging/registry/app-context scoping reads `generated_app_id` via `_shared/build_identity.py`.
2. **Mid-journey transition context loss.** `coding_journey_selector` / `database_setup_selector` dropped journey context to the next workflow. Fix: `SessionState.pending_transition_context` carries scalar journey context across a transition hop.
3. **DesignDocsBundle provider schema.** `DataContractIndex.keys: list[list]` compiled to `items` without a `type`, which OpenAI rejects in strict `response_format`. Fix: typed `DataContractIndexKey`; the strict-schema guard now also rejects bare `list`/`dict`/`Any`.
4. **Sentinel completion fragility** and **5. launch validation fail-closed too hard** — see commit `4b8c83b1`.

## 2. The structured-output / artifact-persistence defect (root cause corrected, now closed)

The earlier draft of this description attributed the missing artifacts to the DesignDocs structured result *"not being present as `structured_output` at auto-tool time on the resumed run."* **That diagnosis was wrong**, and the corrected root cause is narrower and generic:

The validated payload **was always present and exact in value**. The runtime's read-only projection served it through `freeze()`, which rewrites its **shape**:

```
dict -> types.MappingProxyType    (isinstance(x, dict) is False)
list -> tuple                     (isinstance(x, list)  is False)
```

Every consumer guarding extraction with `isinstance(raw, dict)` / `isinstance(raw, list)` therefore **discarded a schema-valid agent output and returned before its persistence work — while still reporting success**, because a tool's early return is a normal return. Nothing was lost at a pause/resume boundary; the failure is unconditional and reproduces on the very first turn.

**Classification: generic runtime boundary defect, not workflow-specific.** The reproduction drives the *unmodified, checked-in* DesignDocs tool, and repairing only `StructuredOutputOverlay.get` fixes it — so no save tool needs a frozen-aware intake shim.

Fix (`0bd927f3`, one call site): the projection returns a fresh **plain `dict`/`list` deep copy per read** of the overlay's own private detached payload. Canonical state stays unreachable without altering the validated result's shape. The #491 projection guarantees are unchanged — writes still fail closed, and the key stays out of snapshots, enumeration, and persistence.

**Why existing coverage missed it:** `tests/test_factory_auto_tool_acceptance.py` exercises ThemeCapture, whose `save_captured_theme` guards on truthiness (`if not data`) — and a `mappingproxy` is truthy. Only the `isinstance` consumers (DesignDocs, SubscriptionContractDesigner) were affected, and none had an acceptance test.

### Live A/B proof (real OpenAI + real Mongo, identical prompt, identical harness)

One bounded greenfield DesignDocs run per arm via `scripts/run_live_workflow_smoke.py`, each with its own disposable `generated_app_id`:

| Arm | Overlay | Run result | `DesignDocuments` persisted under `generated_app_id` |
|---|---|---|---|
| Pre-fix | `freeze()` | exit 0, "success", structured output produced | **0** |
| Post-fix | `detach()` | exit 0, structured output produced | **4** (`frontend`, `backend`, `database`, `ui_schema`, all `status=succeeded`) |

The pre-fix arm reproduces the reported symptom exactly: a clean, successful-looking run that persists nothing. Post-fix, a fresh read returns all four canonical docs scoped to `generated_app_id` (`gen-dogfood-602ee9dae6`), **not** to the executing host identity (`live-smoke-046a7a0d`) — so the #507 execution-identity invariant and the `generated_app_id` target authority both hold through real persistence.

### Upstream artifacts persisting under `generated_app_id`

A live ThemeCapture run against the same generated target (`gen-dogfood-602ee9dae6`) then advanced the journey, which correctly blocked the next workflow (`DesignDocs requires ValueEngine to be completed first`) because this bounded target never ran ValueEngine. Mongo state for that single generated target across both runs:

| Collection | Docs | Meaning |
|---|---|---|
| `ThemeCaptures` | 1 | theme artifact (`theme_config`, `identity`) |
| `DesignDocuments` | 4 | canonical design-doc bundle |
| `DataContracts` | 1 | `status: succeeded` |
| `ArtifactVersions` | 2 | lineage rows, `source_workflow` = `ThemeCapture` and `DesignDocs` |

So theme, design docs, the data contract, and artifact-version lineage all persist under the generated target identity. **Not covered:** subscription contract, workflow bundle, and `app_bundle`.

### Cross-target isolation (design-doc artifact class only)

A third live run under a second generated target, same harness:

| Generated target | Concept | Docs persisted | Contains the *other* target's concept |
|---|---|---|---|
| `gen-dogfood-602ee9dae6` | ShelfList | 4 | 0 |
| `gen-isoB-baca435254` | PlantWater | 4 | 0 |

Each target owns exactly its own four canonical docs, with zero cross-contamination in either direction, and neither is owned by the executing host identity. **Scope note:** this proves isolation for the `design_docs` artifact class only — theme, subscription contract, workflow artifacts, and `app_bundle` are *not* covered by this evidence.

## 3. Build-journey honesty fixes (`d5ac85d0`)

Landed alongside, from live-Genesis forensics: `no_transition_matched` terminations are now **failed** runs rather than clean completions (ThemeCapture and AgentGenerator both terminated on graph exhaustion and the sequence silently skipped their generation stages); ThemeCapture/AgentGenerator gained user-resume and explicit completion routes; fixed-value text triggers classify as deterministic writers so HITL approval flags actually route; a tool returning `ok/success: False` is logged as a failed result instead of "completed successfully"; reused journey chats refresh inherited context; WS clients may answer heartbeat pings.

## 4. This PR's own CI regressions (`7cf1c733`)

The exact-head CI run surfaced two defects that no local suite caught, both introduced by `4b8c83b1`'s identity migration:

- **`lint` (mypy) failed.** `resolve_generated_app_id` returns `str | None`, and both GitHub export call sites passed it into an `app_id: str` parameter. Every other use in those files goes through `str(...)` — which would have published the export under the literal string `"None"` rather than failing. Both sites now **fail closed** on a missing identity: AppGenerator folds it into the existing export gate (reusing the established blocked path), AgentGenerator adds a branch beside its bundle-path guard.
- **`TestAgentGeneratorEmitBuildCompleted` (2 tests) asserted the pre-migration persist signature.** The behavior is correct — `_persist_workflow_bundle_artifact` now also receives `chat_scope_app_id`, which *is* the identity split (`app_id` scopes the artifact to the generated app; `chat_scope_app_id` scopes the chat-context read to the executing runtime app). Only the assertion was stale. Confirmed by running those tests against `origin/main`'s `build_lifecycle.py` (4 passed) vs. this branch's (2 failed).

### `test-shard (3)`: two ValueEngine tests red since this PR's first commit (`d124fc27`)

The exact-head CI run also failed on `tests/test_value_engine_app_record_lifecycle.py`. Bisected: both tests fail identically at **`4b8c83b1`** (this PR's first commit) and pass on `origin/main` — so #521 had been red on **every head since its first commit**, unnoticed by both lanes working this branch.

Same class as the `build_lifecycle` staleness: the tests still assumed `app_id` carries the generated app's registry identity. The behavior is correct — `create_app_record` publishes the generated identity as `generated_app_id` and never returns or rebinds the executing `app_id`.

Reopen safety was explicitly re-verified rather than assumed: an existing `generated_app_id` is reused whenever a `build_registry_id` is present, and the provisional fallback is a **deterministic** digest of `chat_app_id:chat_id`, so a reopened build cannot split its artifact lineage across two identities.

Both tests are updated to the current contract and **strengthened** — each now asserts that the executing `app_id` is still the factory session after the hook runs, which is the #507/#521 invariant the old assertions could not express. Verified non-vacuous: against `origin/main`'s `create_app_record` the two updated tests fail; against this branch's they pass.

## 5. Verification

- Focused suites (combined head `d5ac85d0`): **252 passed** — `test_structured_output_exact_shape`, `test_auto_tool_structured_output_contract`, `test_factory_auto_tool_acceptance`, `test_build_identity_migration`, `test_lifecycle_identity_guard`, `test_structured_output_helpers`, `test_workflow_context_authority_compile`, `test_context_authority_policy`, `test_genesis_persistence_rootcauses`, `test_ws_inbound_keepalive`, `test_reserved_context_keys`.
- New regression `tests/test_structured_output_exact_shape.py` (6 tests) **fails 4/6 against the pre-fix overlay**, including the real-tool case, so it is a real guard rather than a tautology.
- `ruff check .` — clean. `scripts/governance_guardrails.py` — passed, 0 notices. `git diff --check` — clean. Secret scan of both commits — no matches.
- `mypy mozaiksai/ factory_app/` (CI's scope and flags) — **clean, 617 source files, 0 errors**. The repo-wide CI mypy invocation aborts locally on a `numpy` stub (`Type statement is only supported in Python 3.12 and greater`) because the local interpreter is 3.13 while `[tool.mypy] python_version = "3.11"`; **identical with this branch's changes stashed**, i.e. pre-existing and environment-level.

## 6. Still unverified (why this remains DRAFT)

- **No non-empty Revision 1 yet.** The live proof above is a bounded single-workflow DesignDocs run, not a full journey. AppGenerator input resolution, a non-empty generated file map, SecurityReadiness over real files, and AppReview over a real bundle are **not** demonstrated here.
- **Cross-build isolation is proven only for `design_docs`** (see above), not for theme, subscription, workflow artifacts, or `app_bundle`.
- **No Revision 2 refinement proof.**
- The §3 journey/authority/routing changes in `d5ac85d0` are **not independently verified** by the author of `0bd927f3`; they carry their own tests but no live end-to-end evidence in this description.

Do not mark ready, merge, or enable auto-merge until the above are closed.

---

**Real model usage for all live runs in this description:** 6 calls, 73,093 tokens (`gpt-4.1` ×2, `gpt-4o-mini` ×4), $0.0976 — from the `RuntimeUsageEvents` ledger.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,19 @@ This project follows a practical pre-1.0 changelog format:
 
 ### Fixed
 
+- Greenfield Genesis builds run again under the execution-identity guard. The
+  generated app's registry identity now travels in the declared
+  `generated_app_id` context variable — reserved by ValueEngine's registration
+  hook instead of rebinding the immutable executing `app_id` — and all factory
+  artifact, staging, registry, and app-context scoping reads it. Declarations
+  with an explicit non-immutable `authority_class` (registry pointers such as
+  `build_registry_id`, `chat_app_id`) now govern their own keys at the
+  before-chat identity boundary; core execution identity and credential keys
+  remain non-demotable. Mid-journey transition hops (coding journey and
+  database selectors) now carry scalar journey context to the next workflow
+  launch the same way workflow-to-workflow hops inherit it, so build identity
+  survives the full build sequence.
+
 - Security Readiness now saves findings through permissioned workflow module
   dispatch using the live session principal. Findings retain project/build
   association and cannot overwrite another project's matching scanner rule.
--- a/docs/architecture/mozaiksai/context-variables-complete.md
+++ b/docs/architecture/mozaiksai/context-variables-complete.md
@@ -51,8 +51,17 @@ Mutable preload information can still be populated. The check does not undo
 external side effects already performed by a hook, or replace per-writer mutation
 authorization. A generated app's target identity must be resolved separately by
 the owning Factory/Studio build lifecycle; it must never replace the executing
-app's `app_id`. The existing ValueEngine registration hook and downstream artifact
-scoping still require that migration before a hosted full-build proof is safe.
+app's `app_id`. The factory build pipeline carries it in the declared
+`generated_app_id` context variable, written by ValueEngine's registration hook
+and consumed by artifact, staging, registry, and app-context scoping.
+
+Declarations govern non-core names: a resolved declaration with an explicit
+non-immutable `authority_class` (for example a downstream registry pointer such
+as `build_registry_id`, or a `*_app_id` key naming a different entity than the
+executing app) exempts that key from the name heuristic at this boundary. Core
+execution identity (`app_id`, `user_id`, `chat_id`, `workflow_name`, tenant and
+workspace scope, permissions) and credential-bearing keys can never be demoted
+by declaration, and undeclared keys keep the fail-closed name heuristic.
 
 Source types supported by runtime:
 
--- a/factory_app/workflows/AgentGenerator/context_variables.yaml
+++ b/factory_app/workflows/AgentGenerator/context_variables.yaml
@@ -1,4 +1,17 @@
 definitions:
+  generated_app_id:
+    type: string
+    authority_class: mutable_workflow_state
+    authorization: false
+    description: >
+      Registry identity of the app being generated, reserved at pipeline start
+      by ValueEngine's create_app_record lifecycle tool. Scopes artifact-store
+      records, generated staging paths, app-registry updates, and app-context
+      registration. Distinct from the immutable executing runtime app_id.
+    source:
+      type: state
+      default: null
+
   workflow_bundle_export_outcome:
     type: string
     authority_class: closed_writer_routing_state
@@ -340,7 +353,7 @@ definitions:
       database_name: mozaiksai
       collection: BuilderConcepts
       query_template:
-        app_id: '{{runtime.app_id}}'
+        app_id: '{{generated_app_id||runtime.app_id}}'
       fields:
       - ConceptOverview
   concept_api_endpoints:
@@ -352,7 +365,7 @@ definitions:
       database_name: mozaiksai
       collection: BuilderConcepts
       query_template:
-        app_id: '{{runtime.app_id}}'
+        app_id: '{{generated_app_id||runtime.app_id}}'
       fields:
       - ApiEndpoints
       default: []
@@ -365,7 +378,7 @@ definitions:
       database_name: mozaiksai
       collection: BuilderConcepts
       query_template:
-        app_id: '{{runtime.app_id}}'
+        app_id: '{{generated_app_id||runtime.app_id}}'
       fields:
       - Blueprint
       default: null
@@ -379,7 +392,7 @@ definitions:
       database_name: mozaiksai
       collection: DesignDocuments
       query_template:
-        app_id: '{{runtime.app_id}}'
+        app_id: '{{generated_app_id||runtime.app_id}}'
         kind: backend
         status: succeeded
       fields:
@@ -396,7 +409,7 @@ definitions:
       database_name: mozaiksai
       collection: DesignDocuments
       query_template:
-        app_id: '{{runtime.app_id}}'
+        app_id: '{{generated_app_id||runtime.app_id}}'
         kind: ui_schema
         status: succeeded
       fields:
@@ -413,7 +426,7 @@ definitions:
       database_name: mozaiksai
       collection: DesignDocuments
       query_template:
-        app_id: '{{runtime.app_id}}'
+        app_id: '{{generated_app_id||runtime.app_id}}'
         kind: backend
         status: succeeded
       fields:
@@ -439,7 +452,7 @@ definitions:
       database_name: mozaiksai
       collection: ArtifactVersions
       query_template:
-        app_id: '{{runtime.app_id}}'
+        app_id: '{{generated_app_id||runtime.app_id}}'
         build_family: subscription_contract
         build_key: subscription_contract
         lifecycle_status: current
@@ -485,7 +498,7 @@ definitions:
         agent: InterviewAgent
         ui_hidden: true
         match:
-          regex: (?m)^\s*NEXT\s*$
+          equals: NEXT
   workflow_review_approved:
     type: boolean
     description: >
@@ -570,7 +583,7 @@ definitions:
       database_name: mozaiksai
       collection: ChatSessions
       query_template:
-        app_id: '{{runtime.app_id}}'
+        app_id: '{{generated_app_id||runtime.app_id}}'
         _id: '{{runtime.chat_id}}'
       fields:
       - attachments
--- a/factory_app/workflows/AgentGenerator/tools/generate_and_download.py
+++ b/factory_app/workflows/AgentGenerator/tools/generate_and_download.py
@@ -398,6 +398,13 @@ async def generate_and_download(
         user_id = context_variables.get("user_id")
         build_id = context_variables.get("build_id")
 
+    # Generated staging paths, workflow exports, and workflow_bundle artifact
+    # records scope under the generated app's identity; chat persistence and
+    # transport stay on the executing runtime app_id.
+    from factory_app.workflows._shared.build_identity import resolve_generated_app_id
+
+    generated_app_id = resolve_generated_app_id(context_variables)
+
     wf_logger = get_workflow_logger(workflow_name=(workflow_name or "AgentGenerator"), chat_id=chat_id, app_id=app_id)
     tlog = None
     if _get_tool_logger:  # type: ignore[truthy-function]
@@ -508,7 +515,7 @@ async def generate_and_download(
     # Resolve output directory
     # ------------------------------------------------------------------
     base_generated = _resolve_workflow_output_root(
-        app_id=app_id,
+        app_id=generated_app_id,
         build_id=build_id or chat_id,
     )
     base_generated.mkdir(parents=True, exist_ok=True)
@@ -562,7 +569,7 @@ async def generate_and_download(
             _promote_workflow_to_app_workspace(first_wf_dir, first_wf_name)
 
         await _record_context_and_artifacts(
-            app_id=app_id,
+            app_id=generated_app_id,
             user_id=user_id,
             chat_id=chat_id,
             pack_name=bundle_name,
@@ -627,7 +634,7 @@ async def generate_and_download(
             _promote_workflow_to_app_workspace(first_wf_dir, first_wf_name)
 
         await _record_context_and_artifacts(
-            app_id=app_id,
+            app_id=generated_app_id,
             user_id=user_id,
             chat_id=chat_id,
             pack_name=bundle_name,
@@ -704,7 +711,7 @@ async def generate_and_download(
                 wf_logger.info("🚀 Export to GitHub requested (repo=%s)", repo_name)
                 deployment_result = await export_agent_workflow_to_github(
                     bundle_path=str(zip_bundle_path),
-                    app_id=app_id,
+                    app_id=generated_app_id,
                     repo_name=repo_name,
                     commit_message=commit_message,
                     user_id=user_id,
--- a/factory_app/workflows/AgentGenerator/tools/platform/build_lifecycle.py
+++ b/factory_app/workflows/AgentGenerator/tools/platform/build_lifecycle.py
@@ -109,21 +109,28 @@ async def _persist_workflow_bundle_artifact(
     user_id: str | None,
     workflow_name: str,
     build_mode: str | None,
+    chat_scope_app_id: str | None = None,
     artifact_store: Any | None = None,
     workflow_integration_metadata: dict[str, Any] | None = None,
 ) -> None:
     """Persist a versioned workflow_bundle summary artifact after AgentGenerator completes.
 
-    Pass ``artifact_store`` to direct artifact writes to a specific store instance
-    without touching process-level state.  Pass ``workflow_integration_metadata``
-    to supply the payload directly and skip the MongoDB chat-context read.
+    ``app_id`` scopes the artifact record (the generated app's identity);
+    ``chat_scope_app_id`` scopes the chat-context read (the executing runtime
+    app_id) and defaults to ``app_id``. Pass ``artifact_store`` to direct
+    artifact writes to a specific store instance without touching
+    process-level state. Pass ``workflow_integration_metadata`` to supply the
+    payload directly and skip the MongoDB chat-context read.
     """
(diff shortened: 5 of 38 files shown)


elCaptnCode:
Coordination: the operator has asked me to integrate the independently verified codex/factory-build-target work with current main and run a fresh UI-driven Genesis/refinement. That branch has a real generated CRUD app passing Keycloak/Mongo/browser acceptance, checkpoint 613447fc. I will not modify or merge this draft branch. I am auditing the overlap between its generated_app_id pointer and our typed RunBuildBinding execution_app_id/target_app_id contract before integration; please keep this draft unmerged until that overlap is reconciled. Our reference app and your live run will remain isolated.

elCaptnCode:
## Independent verify-by-running review of `d5ac85d0` — 9 confirmed material findings

Scope: only commit `d5ac85d0` ("make build-journey failures loud and HITL gates satisfiable"). Method: 14 agents — 5 dimension reviewers instructed to run code/tests/graph simulations in the real worktree (not read diffs), then adversarial skeptics per material finding instructed to refute. **Every finding below survived refutation (verdict: confirmed). No finding was refuted.** All focused suites pass (95+65+5+94+5+10+20+102+101+71) — which is itself part of the problem: none of these defects is visible to the current test surface.

### Confirmed material

**1. The HITL approval flag is forged by incidental or negated prose — phase-blind, persisted, never reset.**
`workflow_review_approved` / `workflow_review_revision_requested` (`CLOSED_WRITER_ROUTING_STATE`) became user_text-writable by this commit (proven: `can_write` False→True against the real YAML with pre/post `authority.py`). The binding regex `(?i)\b(approved?|looks good|sequence diagram looks good|proceed with implementation)\b` is an **unanchored search** applied by `apply_user_text` on **every** user input in the chat (`workflow_bridge.py:328/535/660`), persisted to Mongo, and nothing ever writes it back to False. Reproduced with the real runtime matcher: `"This is not approved yet"` → approved=True. `"I do not approve this design"` → approved=True. `"looks good so far"` (during the *interview* phase) → approved=True, persisted, and later silently routes the first review reply straight to `PackBuildCoordinator` — the review gate never happens. Rule order makes approve beat revision when both match (`transition_graph.yaml:36-50`, priority = YAML index per `network_graph.py:176-179`).

**2. `no_transition_matched` → FAILED has a 15-workflow blast radius; resume routes were added to only 2.**
Graph compilation + state-fold simulation over all 15 factory workflows:
- **ExistingAppDiscovery** — no terminate route exists at all; every user reply now hard-fails the run. It can never complete successfully.
- **AppReview** — the promote-or-revise HITL step fails on **any** user reply. This is the build journey's *final* gate; it is now unsatisfiable.
- **AppGenerator** — nine RevertToUser recovery paths, **zero** user-resume routes: every HITL unblock reply now fails the run.
- **BrandAssetGeneratorWorkflow** — iteration loop dies on the first user reply. **ValueEngine** — gap state (interview done, concept not yet presented) fails on reply.
- Journey reaction to FAILED verified: hard halt — no retry, no skip.

Before this commit these replies silently truncated the run as "complete" (also wrong). Now they hard-stop the journey. The commit's direction is right; the coverage is 2/15.

**3. `max_turns` exhaustion still reports a clean completion.** `_resolve_close_status` maps `workflow_failed` and `no_transition_matched` but leaves the third silent-skip close reason unmapped — the smoke alarm has a dead zone.

**4. Journey re-entry refresh (`journey_orchestrator.py`): `$set` of `validated_context` with no protected-field guard.** It is the only durable chat-session write path without one — a context key colliding with a reserved session-doc field (e.g. `status`, `messages`) clobbers the document. Also confirmed: the refresh never reaches an **in-process paused run**, which later re-persists the stale identity over the refreshed doc (the exact split-lineage failure this hunk exists to fix); failure is swallowed to a warning and `matched_count` is never checked.

### Minor (confirmed, listed for the record)
- "Deterministic sentinel" broadening enforces no anchoring/case shape — the commit message's "anchored regex" contract exists nowhere in code; only factory workflows are test-guarded.
- `$1`-capture triggers are misclassified as sentinel in the live derived-manager path (declared value discarded at construction — the commit's `$1` guard is dead code there).
- The new sentinel-subset acceptance in `can_write` skips the routing/authorization guards the AGENT_TEXT branch enforces.
- ThemeCapture user-resume routing has no fallback for a non-bool/missing `interview_complete`; its sentinels also keep the brittle exact-equals form the commit replaced in AgentGenerator for the same failure mode.
- `DownloadAgent` ready→terminate closes the channel immediately after presenting the DownloadCenter inline UI.
- Keepalive whitelist comment/test overstate the server mechanism (idle refresh is independent of validation); authority compile test duplicates runtime writer-resolution logic instead of importing it.

### ⚠️ To the in-flight lane on this branch
There is uncommitted work in the shared worktree right now adding a `(LIVE_USER_CONTEXT_WRITER → USER_TEXT_TRIGGER_WRITER)` elevation in `ag2_network_runner.py` — that builds **additional** trust on the forgeable mechanism in finding 1. Please anchor the approval/revision bindings before extending trust in user_text writes, and land the in-flight work so fixes can proceed without collision.

### Proposed fixes (either lane can execute; smallest honest versions)
1. Anchor both bindings to full-message sentinels (e.g. `(?im)^\s*(approved?|looks good|proceed)\s*[.!]?\s*$`); prefer revision over approval when both match; reset the flags at review-phase entry.
2. Add user-resume routes to ExistingAppDiscovery, AppReview, AppGenerator, BrandAssetGenerator, ValueEngine (pattern: this commit's own AgentGenerator `after_turn` fallback), plus a structural test: **any graph with RevertToUser targets must declare ≥1 user-source resume route** — that guard would have caught all of finding 2.
3. Map `max_turns` → FAILED in `_resolve_close_status`.
4. Strip reserved session-doc fields from the refresh `$set`; check `matched_count`; document the live-session race as a bounded follow-up (not an in-PR redesign).

*Review run: 14 agents, 0 errors, findings adversarially verified; full evidence retained in session records. PR remains draft; nothing here is merged.*

elCaptnCode:
## Review findings closed in `d60b6329`

Addresses the confirmed findings from the review comment above. Every claim below is reproducible from the diff or the pasted results.

| # | Finding | Status |
|---|---|---|
| 1 | HITL approval forgeable by negated/incidental prose | **Fixed** |
| 2 | 4 workflows unusable after any user reply | **Fixed** (all 4) |
| 3 | `max_turns` still reported clean completion | **Fixed** |
| — | No guard against this class returning | **Added** |
| 4 | Journey `$set` reserved-field guard + paused-run race | **Deferred, documented** |

**1 — approval sentinel anchored.** Verified against the real `UserTextBinding` matcher, before vs after:

| Reply | before | after |
|---|---|---|
| `"this is NOT approved"` | approved=True | approved=False |
| `"I do not approve this design"` | approved=True | approved=False |
| `"looks good so far"` | approved=True | approved=False |
| `"It looks good overall, but please rename StepTwoAgent first"` | approved=True | approved=False |
| `"The sequence diagram needs changes but is otherwise approved"` | approved=True | **revision=True** |

**6 forgeries → 0 accepted; 0 of 6 genuine approvals lost** (`approved`, `Approved.`, `looks good`, `Looks good!`, `proceed with implementation`, `APPROVE` all still match). The revision rule is now also evaluated before approval, so a reply that both requests changes and contains an approval word can never read as consent.

**2 — resume routes added to all four graphs.** AppGenerator's mirror each pause condition exactly (`app_ui_quality_status=blocked → AppUIQualityAgent`, `app_bundle_export_outcome=cancelled|blocked|ready → DownloadAgent`, …), so a reply resumes the agent that asked for it rather than a hub guess. ExistingAppDiscovery resumes at the earliest incomplete stage via rule-order priority. AppReview and BrandAssetGenerator are single-agent, so their target is unambiguous.

**3 — confirmed at the source before fixing.** `ag2/network/adapters/workflow.py:335-338` sets `auto_close_reason="max_turns"`; the literal was simply unmapped in `_resolve_close_status`. Now FAILED, same as the other two silent-skip reasons.

**Guard — `tests/test_workflow_user_resume_routes.py`.** Any graph that reverts to the user must declare a user-source resume route, and resume targets must be declared agents. **Verified non-vacuous:** against the pre-fix graphs it fails exactly `AppGenerator`, `AppReview`, `ExistingAppDiscovery`, `BrandAssetGeneratorWorkflow` and passes on the repaired ones.

### Two judgment calls, flagged rather than buried

**A retired contract test was changed, not just satisfied.** `test_appgenerator_handoffs_start_from_agents_not_pseudo_user` banned user-source rules outright. That ban dates from the AG2 beta (#32), is already contradicted by AgentGenerator / ThemeCapture / ValueEngine, and became actively harmful once unmatched replies started failing runs — it made AppGenerator permanently unsatisfiable. It now asserts the property the ban was protecting (*a reply may only resume an agent that pauses for input, never jump into the middle of the generation chain*) instead of forbidding resume routes.

**One state was not guessed at.** On ExistingAppDiscovery's ecosystem / gradual_modernization paths, a reply sent while `ModuleDecomposerAgent` is still working resumes at the assembler, because `decomposition_complete` is also false on embed/bridge (where the decomposer never runs) and `context_equals` cannot test two keys. Routing the tail to the decomposer would mis-route embed/bridge instead. Resolving it properly needs a decomposer-specific resume signal. The limitation is written into the graph; the run completes rather than failing outright.

### Still open (deliberately not in this commit)
- Journey re-entry `$set` reserved-field guard, `matched_count` check, and the in-process-paused-run race — a bounded follow-up, not an in-PR redesign.
- Residual cross-phase persistence of the approval flag: `UserTextBinding` has no phase/condition field (`variable/equals/contains/regex/value` only), so gating the write on review-phase entry would mean inventing schema. Anchoring shrinks the window; it does not close it.
- The minor findings from the review (sentinel anchoring/case contract undocumented, `$1`-capture misclassification in the derived path, ThemeCapture exact-equals brittleness, DownloadAgent terminate-after-UI).

**Verification:** `2703 passed, 23 skipped, 0 failed` on the graph/authority/generator sweep; `147 passed` focused; `ruff check .` clean. PR remains **draft** — not merged, auto-merge not enabled.

elCaptnCode:
Integration PR #524 is now open from current main, with the prior verified Factory checkpoint and fresh normal-UI Contact Desk evidence. It preserves immutable host app_id and the server-owned RunBuildBinding; it does not merge #521 or introduce execution_app_id aliases. Real app CRUD/auth/refinement/download/local promotion passed. Clean local full rerun and Linux CI are running; App Zero installed-wheel acceptance passed and final consumer pin remains gated on integration. Please coordinate on #524 before rebasing overlapping Factory identity changes.

PR #533, opened 2026-09-13
fix(factory): ground workflow prompts in semantic inputs

elCaptnCode:
Factory prompts sometimes asked models to consume outputs by another agent's name or through a wrapper that was absent from the recipient's prompt. This change describes upstream inputs by their actual fields, semantic meaning, authority, and effect on the next output. For example, workflow-plan review now consumes the exposed `workflows_spec` and preserves its proposed topology, while research findings inform concept synthesis without becoming user approval.

The same rule now applies to first-party prompts, injected catalogs/hooks, and generated workflow guidance. Contributor guidance remains canonical in `docs/agent-engineering-contract.md`; `CLAUDE.md` imports it explicitly, and the entrypoints reference shared rules instead of repeating them. Machine agent identifiers remain in dispatch, roster, tool bindings, and registries. All 15 workflow rosters/settings and structured-output shapes are unchanged.

Validation:

- `ruff check .`: all checks passed.
- Focused 12-file generator, prompt, hook, and runtime-smoke suite: 259 passed, 1 warning in 24.39s using Python 3.13 and the required AG2 1.0.3.
- Guidance checks: 17 passed, 1 warning; 20 relative links/anchors resolved.
- All changed YAML parses; `git diff --check` is clean.
- Full repository suite (`python -u -m pytest -q --no-cov`): 17,430 passed, 7 failed, 102 skipped, 5 warnings, 14 reruns in 1812.75s. All seven failures were stale wording assertions; each has been corrected and rerun (7 passed, 1 warning in 4.03s).
- Final 10-file prompt/catalog/hook regression run, including all seven failed tests and the final self-contained proposal instruction: 367 passed, 2 skipped, 1 warning in 14.68s. The full repository suite was not repeated after these fixes.

OSS Change Impact: Factory prompt behavior and contributor documentation change; runtime/platform composition, app workspace shapes, and workflow sequencing do not change.

AgentGenerator Workflow Impact: corrected prompt inputs, universal injection, schema descriptions, and generated-agent authoring guidance. No output schema migration or routing change. Coverage includes real context projection and deterministic generated-workflow loading/routing/task execution; no live LLM generation was run.

Existing limitations: ThemeCapture and ValueEngine intermediate narrative briefs are usable only when supplied in the current messages; this change does not introduce a durable handoff. The workflow export tool still consumes completed bundle results rather than a separately exposed pack-metadata artifact, so the download prompt no longer claims to inspect that unavailable artifact.

Base: `d5e7e13fab23449fa9399db76e6f8a62d9ec101f`.


elCaptnCode:
Coordination: #531 is an earlier in-flight Factory live-acceptance fix and overlaps AppGenerator agents/hooks and SecurityReadiness agents. It contains tested canonical filename/action binding guidance and zero-findings recording semantics. Please preserve those behavioral fixes when reconciling prompt edits. #531 remains draft while real generated-app revision acceptance runs; do not stack either branch on the other unmerged PR.

elCaptnCode:
Reviewed, rebased onto `main`, and pushed two follow-up fixes (commit `7484bff7`).

**Review outcome:** sound change — every Python edit is a string literal (no logic changed), and I verified all seven reworded test assertions track the prompts faithfully rather than weakening them (two files actually gained assertions). Two things needed attention:

1. **Restored the exact-identifier examples.** `SEMANTIC_REFERENCE_RULES` is labeled `CRITICAL - PREVENTS CROSS-REFERENCE ERRORS`, and the rewrite replaced its worked ✅/❌ pairs and the causal "why" ("the runtime performs exact string matching") with abstract prose — concrete-example density in that file dropped 13 → 5. Verbatim identifier copying is the failure mode most helped by worked examples and most hurt by abstraction. I restored them **rephrased to name the supplied artifact rather than a producing peer**, so the semantic-input discipline this PR introduces is fully preserved (the new gates pass on the restored text).

2. **Exempted two declarative shapes in the new prose guard.** The exemption regex covers keyed machine fields but not two shapes the workflow contract actually uses:
   - `ui_config.yaml` → `visual_agents:` bare list members
   - `structured_outputs.yaml` → `registry:` mapping keys (agent id as key)

   Both were flagged as reasoning dependencies. Verified by probing the helper directly. The suite is green today only because current prompts use placeholder names in examples — the first prompt showing a real bundle example would fail confusingly. Both shapes are now exempt and pinned, along with a bare in-prose identifier that must still flag.

   Also made the roster read defensive so an `agents.yaml` without an `agents` key fails as a test rather than a collection-time `KeyError`.

**Verification:** full suite on the rebased branch — **17,506 passed, 99 skipped**; `ruff check .` clean; the 11 PR-touched suites pass at 403 passed / 2 skipped.

**Note for a follow-up, not blocking:** this PR changes model behavior across 15 rosters and is validated entirely by string assertions — the description notes no live LLM generation was run. Worth one real generation run against the new prompts when convenient, per the active verification-priority rule.

Issue #576, opened 2026-09-14
Duplicate websocket for a chat evicts the live socket and kills the running workflow


elCaptnCode:
Found by a live traversal on `468c9c08`. `AgentGenerator` died 11 seconds into a real build, and the UI showed **"Retry failed workflow"**.

## Evidence chain

```
17:20:58.860  [ORCHESTRATION] Starting AgentGenerator workflow
17:20:59.095  WORKFLOW_STARTED chat_id=2b097215… mode=AgentDriven
17:20:59.289  CHAT_LOCK_BUSY resource=chat:mozaiks-factory:2b097215… local_holder=true
17:21:00.068  Error: This chat is already executing elsewhere. Please retry shortly.
17:21:00.079  Evicting stale WebSocket for chat_id=2b097215… (ws_id=1736695082224)
17:21:00.080  Workflow 2b097215… already exists in session …, updating instead
17:21:00.028  Unexpected ASGI message 'websocket.send', after sending 'websocket.close'
17:21:09.841  [AGENTGENERATOR] Run failed: workflow_failed  duration=11.11s  agents=7
```

Note `local_holder=true` — this is self-contention inside a single process, not two servers.

## What is NOT the bug

Two paths I suspected and cleared, so nobody re-walks them:

- `_reject_chat_locked` (`workflow_bridge.py:446`) is correct. It logs, notifies, returns `{"status": "busy"}`, and explicitly fails closed *before* any session/WAL mutation. It does not touch the in-flight run.
- `_flush_message_queue` (`ws_protocol.py`) already catches send failures, and drops the queue when the chat is gone.

## What is the bug

`handle_websocket` (`simple_transport.py:1609`) evicts a stale connection when a second websocket arrives for the same `chat_id`:

```python
if chat_id in self.connections:
    stale = self.connections[chat_id]
    await stale_ws.close(code=1001)
    await self._cleanup_connection(chat_id)
```

The eviction is deliberate and its comment is sound — without it the old receive loop's `finally` deletes the new connection entry. **But it does not consider that a workflow may be actively executing on that chat.** The running build's output stream is closed underneath it, and the run fails.

## Impact

A user whose browser reconnects, double-navigates, or races a sequence transition loses an in-flight build. It presents as a generic "workflow failed", so the real cause is invisible from the UI.

This looks like the residual fencing gap noted in #426.

## Not obvious which fix is right

- eviction defers while an execution holds the chat lease, or
[...]

Issue #426, opened 2026-08-29
scale: execution fabric gaps — dead distributed lock, unwired durable queue, unreachable cancel, no orphan reconciliation


elCaptnCode:
Found by the AG2 reuse & scale-readiness audit (main @ f6a8b34c). Four coupled gaps that together block >1-instance and high-concurrency operation:

1. **The distributed chat lock is dead code.** `mozaiksai/core/runtime/persistence/distributed_lock.py:156` is defined, its TTL indexes are created at startup (`hosts/runtime.py:359`), and comments at `hosts/runtime.py:1158-1163` + `persistence_manager.py:849` claim it is the mutual-exclusion boundary — but there are **zero `distributed_lock(...)` call sites**. Two instances can resume the same chat and interleave session writes and WAL sequence allocation.

2. **A real durable queue exists and is unwired.** `mozaiksai/core/workflow/queue.py` (leases, fencing, dead-letter, max_attempts) defaults to `noop` and its only consumer is the task-batch `work_assignment_worker`. The chat/build spawn path never enqueues; admission control is a per-process `Semaphore(MOZAIKS_MAX_PARALLEL_WORKFLOWS=4)` inside the `SimpleTransport` singleton (`simple_transport.py:177`), with no per-tenant fairness and unbounded pending tasks.

3. **Cancellation is unreachable.** The only cancel path (`AG2OrchestrationAdapter.cancel()` → `pause_background_workflow`, `workflow_bridge.py:839`) has **no production caller** — no stop/abort HTTP or WS endpoint exists. WS disconnect is a no-op on compute; spawned side effects (sandbox sessions, docker/git subprocesses) have no kill-on-cancel wiring.

4. **No orphan reconciliation.** On process crash/redeploy, in-flight runs stay `IN_PROGRESS` in Mongo forever: `_background_task_failure_callback` only logs (`workflow_handlers.py:23-39`) and early-returns on cancelled; the post-run persistence block in `orchestration_patterns.py:1417-1455` is skipped on CancelledError; `_runtime_shutdown` drains nothing. Admin dashboards and session-router active queries key off these stuck records.

Remediation direction (mostly wiring, not new infrastructure): acquire `chat:{chat_id}` around run/resume with lease renewal; route spawns through MongoWorkflowQueue with per-tenant keys; add a stop endpoint that calls the existing adapter cancel + tears down a run-scoped resource registry; startup sweep marking stale IN_PROGRESS runs resumable.

[...]

PR #598, opened 2026-09-15
fix(transport): releasing a connection alias must not close the socket

elCaptnCode:
The collision behind #576, and the reason #586 did not remove the race.

## It was never a duplicate socket

`journey_orchestrator._ensure_connection_alias` registers the **source chat's websocket under the next chat's id**, so events for the next stage reach the user before their client reconnects. The name says it: it is an *alias*, not a second connection.

Meanwhile the client sees a new `chat_id` in the route and connects its own socket — correctly. `handle_websocket` then evicts whatever occupies the slot:

```python
stale_ws = stale.get("websocket")
await stale_ws.close(code=1001)     # ← the socket the user is still on
```

For an alias, that closes a live connection in the middle of a build. A real traversal lost `AgentGenerator` to it: **4 evictions, 4 `LockAcquisitionError`s in one run**.

This also settles an ambiguity from my first report on #576. One `ws_id` appeared under two chats and I could not tell whether that was one socket re-keyed or two sockets sharing a recycled address. It was one socket, aliased on purpose.

## The fix

Alias entries carry the chat that owns the socket. The eviction releases the slot and leaves the socket alone. A genuinely stale socket is still closed, so the eviction keeps doing its job.

Both mechanisms were already right — the alias covers the reconnect gap, the client follows the route. Only the eviction could not tell them apart.

## Verification

| check | result |
|---|---|
| new `test_connection_alias_is_not_closed` | **0/4 on unfixed main, 4/4 after** |
| transport / websocket / journey / session / chat-lock | 603 passed, 15 skipped |
| `ruff check .` | clean |

Two of the four tests pin the marker **at both ends** — one asserts the orchestrator sets it, one asserts the eviction reads it. A label nobody checks, or a check for a label nobody sets, would otherwise pass silently.

I also extracted the decision into `_is_connection_alias` so the test calls the real predicate. My first version tested a copy of the logic and passed against unfixed main, which is the kind of test that certifies nothing.

## Scope

This removes the failure. It does not restructure connection ownership — `connections[chat_id]` still carries both real connections and aliases, and #426's phantom entries from `handle_switch_workflow`/`handle_start_workflow` are untouched. Those remain open questions there.


elCaptnCode:
Superseded by #603, which includes the alias fix plus the missing source_chat_id test update that caused CI failures.

PR #636, opened 2026-09-16
fix(pages): point endpoints at the module that declares the action

elCaptnCode:
## The defect

A generated habit tracker emitted:

```
/api/modules/habits/create_habit
/api/modules/habits/checkoff_habit
```

against a module whose id is **`habits_registry`**. Every call 404s; acceptance reports both actions as orphaned.

## Why this is not a prompt fix

The page agent is told the canonical form **four separate times**, including the exact anti-pattern:

> `api_endpoint` … must be a canonical module action path using the exact generated action id: `/api/modules/{module_id}/{action_id}`
> Copy every module action id exactly from dependency module.yaml outputs
> **Do not preserve a nonexistent API path to match design docs.**

It followed that rule. The **shape** is right — it invented the **identity**. A fifth statement of a rule already stated four times would not have helped, and I'd have concluded "prompts don't work" from a prompt that was already correct.

So this is a deterministic rewrite against the emitted contracts, applied in the existing page-normalization pass where the full `file_map` (module.yaml included) is already in hand.

## Deliberately narrow

| Case | Behaviour | Why |
|---|---|---|
| unknown module, **exactly one** module declares the action | retargeted | the only unambiguous reading |
| unknown module, **two** modules declare it | left alone | picking one silently binds a module the planner may not have meant |
| unknown module, **no** module declares it | left alone | nothing to point at |
| **real** module, action missing | left alone | `module.yaml` is the contract, not this |
| `/api/habits` (non-canonical) | left alone | rewriting means *inventing* an action id |

That last row is the read-action defect (#634), not this one. The module gaining a read is what removes the reason the page wrote it — repairing the symptom here would paper over it.

## Verified against the real bundle

Run against the emitted files from the run that produced the defect, not only fixtures:

```
module index: {'habits_registry': ['checkoff_habit', 'create_habit']}

ui/pages/create.yaml
  before: ['/api/modules/habits/create_habit']
  after : ['/api/modules/habits_registry/create_habit']

ui/pages/dashboard.yaml
  before: ['/api/habits', '/api/habits/streaks', '/api/modules/habits/checkoff_habit']
  after : ['/api/habits', '/api/habits/streaks', '/api/modules/habits_registry/checkoff_habit']
```

Both canonical endpoints corrected; both non-canonical ones untouched, as intended.

A test also covers endpoints buried inside a JSON-encoded `config_hint` string, since table endpoints live there rather than in a YAML field.

## Verification

New suite 11/11 · `pytest -k "page or task_batch or assemble or appgenerator"` → **1817 passed, 21 skipped** · `ruff check .` clean.


elCaptnCode:
## Proof against the real acceptance check

Ran `validate_wiring` — the actual `module_action_wiring` check — over the emitted bundle, before and after this change.

**Before:**
```
passed: False   wired: 0   orphaned: 4
  create/habit-creation-form/submit_action -> /api/modules/habits/create_habit
  dashboard/habit-list                     -> /api/habits
  dashboard/habit-list/checkoff_habit      -> /api/modules/habits/checkoff_habit
  dashboard/streak-dashboard               -> /api/habits/streaks
```

**After:**
```
passed: False   wired: 2   orphaned: 2
  wired:    /api/modules/habits_registry/create_habit
            /api/modules/habits_registry/checkoff_habit
  orphaned: /api/habits
            /api/habits/streaks
```

Both canonical-form endpoints resolve. The two left are exactly the non-canonical ones this change deliberately does not touch.

## What this does not fix, stated plainly

The check still fails, and this PR alone will not make it pass.

- `/api/habits` should be closed by #634 — once the module declares `list_habits`, the page has a real read to bind to instead of inventing one. **Plausible, not yet proven**; it needs a run with #634 active, and I will not call it fixed before seeing that bundle.
- `/api/habits/streaks` is a *second* read, not the same one. #634 synthesizes a single `list_*` per module, so this one likely survives both fixes. Closing it means the planner declaring a streaks operation — a design decision, not something a repair should invent.

I would rather name that gap now than have it turn up as a surprise in the next run.

PR #664, opened 2026-09-18
test(traversal): land the live traversal harness

elCaptnCode:
## What

The harness that drives the real product loop — idea to generated bundle — through a running Studio, with real LLM calls and real Mongo. Nothing mocked.

It found both `AppGenerator` defects fixed this week (#653, #658) and existed **only on one machine, untracked**. Nobody else could run it, so nobody could reproduce anything it reported, and one `git clean` would have deleted it.

## Why it earns its place

A mocked test of an unproven seam proves the mock. This drives the actual stack, which is the only reason those two defects were found at all — both were invisible to the existing suite and only appeared in a live run.

## The verdicts are the point

```
BUNDLE PRODUCED      a chat THIS RUN drove produced files
NO BUNDLE            ran, did not fail, produced nothing
WORKFLOW FAILED      a workflow died; the backend log has the reason
BLOCKED              never ran — e.g. the token gate rejected it
UNKNOWN              the probe could not answer — not a negative
```

These used to collapse into one line. That is how a run which **never started** (exhausted token wallet, 400 at the door) was reported as a build that produced nothing — 40 minutes spent, and the conclusion was about the wallet, not the product.

## Traps it guards, each of which cost a run

- **Counting bundles globally.** Other agents build against the same Mongo. A concurrent bundle raised the count and the harness reported success — for an app it had not built. It attributes by `chat_id` now.
- **Watching `generated/apps`.** Generated code lives in `ChatSessions.generated_files`. That directory stays empty through a *successful* build.
- **Measuring `body.innerText`.** Artifact panels are outside body text, so a live screen read as a stall.
- **Anchored gate labels.** Option cards concatenate badge + title, so the accessible name can begin `FASTEST`, not the title.
- **Treating a reply as progress.** A dialog awaiting a click is indistinguishable from a healthy interview by DOM alone — hence `REPLY_LOOP`, which also dumps every visible button so an unrecognised gate names itself.
- **Retiring a gate permanently.** A healthy `Continue` looked inert twice and was abandoned; the run then replied for 20 minutes with the button on screen. Exhaustion now decays (`GATE_FORGIVEN`).

## Changes for portability

`count_generated.py` no longer hardcodes an absolute `.env` path — it reads `MONGO_URI` from the environment or the repo `.env`, so it runs on any checkout. Verified against live Mongo (`{"bundles": 7, ...}`).

Added `npm run traversal`. Artifacts are gitignored: ~7MB of screenshots per run.

## How to run it

It drives an already-running stack and starts nothing itself. Three prerequisites, each of which has silently produced a false result before — including a funded wallet for `anonymous` (the identity the **backend** sees, not the `dev-user` the frontend displays). The README covers them.

## Not claimed

This does not run in CI and should not: it needs a live Studio, real model calls, and a funded wallet. It is a developer instrument for proving the loop, not a gate.


elCaptnCode:
Superseded by a branch rebased onto current main with the source-hygiene violation fixed. The gate was correct to reject it: this file used "backward" in a navigational sense, but the scan bans that word to keep compatibility-shim language out of the tree and cannot distinguish the two. Renamed to REVERSING and reworded the comment rather than touching the gate.

PR #667, opened 2026-09-18
test(traversal): land the live traversal harness

elCaptnCode:
Replaces #664, rebased onto current `main` with the CI failure fixed.

## Why #664 failed, and why the gate was right

```
Source hygiene scan failed:
  web_shell/live-traversal/traversal.spec.js:107: // and backward options are never listed.
tests/test_production_readiness_gate.py::test_source_hygiene_scan_passes_current_repo FAILED
```

The scan bans `backward`/`backwards` to keep compatibility-shim language out of the tree. This file used the word in a **navigational** sense — options that undo or cancel rather than advance — but the scanner cannot distinguish that from a versioning claim, and it is right not to try.

Fixed by saying what was actually meant: the set is now `REVERSING`, and the comment reads "options that undo or cancel are never listed." **The gate is untouched** — weakening a release gate to admit a test harness would be the wrong trade.

Verified locally, both the scan and the test that guards it:

```
violations: 0
tests/test_production_readiness_gate.py  11 passed, 1 skipped
```

That one word was the cause of all three failing checks: `lint`, `test-shard (2)`, and the aggregate `test`.

## What this lands

The harness that drives the real product loop — idea to generated bundle — through a running Studio with real LLM calls and real Mongo. It found both `AppGenerator` defects fixed this week (#653, #658) and existed **only on one machine, untracked**; nobody else could reproduce anything it reported.

Its verdicts are the point:

```
BUNDLE PRODUCED      a chat THIS RUN drove produced files
NO BUNDLE            ran, did not fail, produced nothing
WORKFLOW FAILED      a workflow died; the backend log has the reason
BLOCKED              never ran — e.g. the token gate rejected it
UNKNOWN              the probe could not answer — not a negative
```

These used to collapse into one line. That is how a run which never started (exhausted token wallet, 400 at the door) was reported as a build that produced nothing.

Traps it guards, each of which cost a real run: counting bundles globally (another agent's bundle was reported as this run's success), watching `generated/apps` (generated code lives in Mongo; that directory stays empty through a *successful* build), measuring `body.innerText` (artifact panels sit outside it, so a live screen read as a stall), anchoring gate labels (option cards concatenate badge + title, so the name can begin `FASTEST`), and treating a reply as progress.

## Not claimed

This does not run in CI and should not: it needs a live Studio, real model calls, and a funded wallet. It is a developer instrument for proving the loop, not a gate.


elCaptnCode:
Coordination note — there are now two live traversal harnesses in this repo, and one hard-won finding that will bite this one.

## The finding: writing under `web_shell/` reloads the app you are driving

My harness (`web_shell/playwright/live.tail.spec.js`, #631) wrote its screenshots and report to `web_shell/live-tail/`. Vite watches `web_shell`. The websocket showed it plainly:

```json
{"type":"full-reload","triggeredBy":"...\web_shell\live-tail\tail-report.md","path":"*"}
```

Every screenshot reloaded the application under test. After the reload the composer stopped accepting input and the run died — and I diagnosed it as a **product-side stall**, even built a hypothesis about a non-terminating interview on top of it, before the `full-reload` frame naming its own trigger file made it unambiguous.

#640 moved output to `.local/live-tail`, outside the watch root, with a `TAIL_OUT_DIR` override. `web_shell/live-traversal/` is a sibling of the directory that caused this, so I would expect the same behaviour. Worth checking before you spend a run on it.

## The duplication

- `web_shell/playwright/live.tail.spec.js` + `playwright.tail.config.js` (mine, merged in #631)
- `web_shell/live-traversal/traversal.spec.js` + `traversal.config.js` (this PR)

Two harnesses for one job is waste, and yours is better instrumented for measurement — `count_generated.py` attributing bundles to chat ids is exactly what I did by hand across seven runs today. `REPLY_LOOP`/`UNMATCHED_BUTTON` and the immediate `WORKFLOW_FAILED` stop are also independent rediscoveries of things mine learned the slow way.

I am not blocking this. My suggestion is yours becomes canonical and mine is deleted, provided it carries these three lessons:

1. **Output outside the vite watch root** (above).
2. **Verify a typed reply was actually delivered.** Playwright resolves `fill`/`press` against a disabled, readonly, or mid-reload composer without sending anything. Mine logged four nudges while `chat.input_ack` stayed frozen at 10 — four turns it reported and never took. Checking that the composer cleared is what caught it (#631).
3. **Stop after N undelivered replies with the real reason**, rather than spinning to timeout. That alone turned a 1.3h dead run into a 13 minute one with an accurate diagnosis.

If you would rather keep both, say so and I will leave mine in place — but then it should be clear which one is authoritative, because I have been using mine to produce the success-rate numbers.

Happy to delete mine once this lands.

PR #680, opened 2026-09-20
fix(app-generator): the tools that set task batch status may write it

elCaptnCode:
## The defect

PR #678 gave `app_task_batch_status` an explicit writer list:

```yaml
app_task_batch_status:
  authority_class: closed_writer_quality_state
  writer_ids: [task_batch]
```

An explicit `writer_ids` **replaces** the inferred set rather than extending it ([`infer_context_authority`](https://github.com/BlocUnited-LLC/mozaiks/blob/main/mozaiksai/core/workflow/context/authority.py#L497-L501)), so `deterministic_tool` — the attribution every declared workflow tool runs under — dropped off the key.

Three tools write it: `app_build_plan.py:1881`, `:2072`, and `app_plan_review.py:28`. The last one is the **opening statement** of `review_app_build_plan`, before any validation:

```python
def review_app_build_plan(*, AppBuildPlan=None, context_variables=None):
    ...
    _clear_plan(context_variables)     # <- rejected here
```

So AppGenerator failed on its first plan review of every run, and no plan ever reached a validator.

## Reproduction

Against the real contract at `2f44874a`, replaying `_clear_plan` through a correctly scoped, **bound** tool invocation:

```
pre-#678   writers = [deterministic_tool, lifecycle_tool, structured_output, task_batch]   -> OK
at  #678   writers = [task_batch]                                                          -> REJECT
```

```
context_authority.rejected workflow=AppGenerator key=app_task_batch_status writer=context_bridge op=set
```

`_clear_plan` writes four keys. The three before this one are `mutable_workflow_state` and land — which is why a failed run persists `app_build_plan: null` and `app_task_batch_items: []`, and the batch then reports "produced no task items".

### The error message points at the wrong thing

`resolve_declared_context_writer` only honours a declared writer that is **already in** `writer_ids`; otherwise it falls back to `context_bridge` *before* the authority check. So a writer missing from the list is reported as `writer=context_bridge`, which reads like the bridge overstepping rather than the declaration being too narrow. The bridge never tried to write. This cost real diagnosis time and is worth knowing next time this message appears.

## The change

One line (plus a comment recording the above):

```yaml
writer_ids: [deterministic_tool, task_batch]
```

Both are real writers: the tools set `planned`, and the batch runner owns the key as its `status_key` (`task_batches.yaml:40`). The result is **strictly narrower than the pre-#678 inferred set** — `lifecycle_tool` and `structured_output` are not restored:

```
ALLOWED: deterministic_tool, task_batch, runtime_system
DENIED : context_bridge, agent_text, persisted_replay, caller_input, live_user_context,
         lifecycle_tool, sentinel_text_trigger, structured_output, tool_writeback,
         transition_router, ui_response_trigger, user_text_trigger
```

`app_task_batch_results` was narrowed by #678 too, but **no tool writes it** — only the batch runner's `context_key`. It is correct as-is, left untouched, and pinned that way by a test so this fix is not mirrored onto it.

## Why CI did not catch it

`test_workflow_context_authority_compile.py` asks whether a routing key has *some* deterministic writer. `task_batch` is one, so the guard passed while the tools were locked out. The new test asserts the declaration covers the writers that **actually write** the key.

## Tests

`tests/test_appgenerator_task_batch_status_writers.py` (new, 17 cases incl. a parametrized deny sweep):

- `review_app_build_plan` can clear and write the key — the live failure, exercised through a real bound tool invocation
- `deterministic_tool` and `task_batch` can each advance it
- `context_bridge` stays unauthorized, asserted on the unbound-bridge shape that produced the live rejection
- every other writer stays rejected (parametrized over `_ALL_WRITERS`)
- the declaration covers every tool that writes the key — the guard the compile-time check cannot give
- `app_task_batch_results` stays batch-only

Verified non-vacuous by reverting the one-line fix and re-running:

```
without the fix:   3 failed, 14 passed
with the fix:     17 passed
```

The three that flip are exactly the regression; the `context_bridge` and unrelated-writer tests pass in **both** states, which is the point — they assert the fix changes nothing about the security property.

```
focused (authority / appgenerator / context_variable / task_batch /
  workflow_context / plan_review / app_build_plan / app_plan)
                                    2332 passed, 17 skipped
ruff check .                        All checks passed!
source hygiene scan                 violations: 0
```

The full suite was still running locally when this PR opened; CI runs it here.

## Scope

Three files: the one-line contract change, the new test, and a CHANGELOG entry. No change to `authority.py`, runtime code, `app_task_batch_results`, App Zero, or any global authority behavior.

## Not claimed

This removes the rejection that ended the run at plan review. It does not prove a build now completes — no live acceptance has been run against it, and that is deliberately held until this merges.


elCaptnCode:
The full local suite finished after this PR opened — clean, no regressions:

```
pytest -q --no-cov     18503 passed, 118 skipped, 0 failed     (17:54)
```

That closes the one gap the description flagged ("the full suite was still running locally when this PR opened").

PR #682, opened 2026-09-20
fix(app-generator): show the planner the subscription config contract

elCaptnCode:
## The defect

`AppBuildTask.owned_paths` is documented as *"copy exact paths from FILE CONTRACTS CONTEXT."* That context is assembled from `_PLANNING_CONTRACT_ORDER`, which listed **six** of the **seven** task contracts in `file_contracts.yaml`. `subscription_config` was the one left out.

The planner was still told `subscription_config` was legal vocabulary, so it planned the task and had no path to give it. A live greenfield run (`build_314a761d`, OSS `50270b84`) emitted `owned_paths: []` three times and died with its whole attempt budget spent:

```
Build task '6' uses task_type 'subscription_config' but owns [].
Subscription config tasks may only own config/subscriptions.yaml.
```

## Reproduced offline against the real contract

```
task_contracts in data : api_surface, module_contract, page_bundle, persistence_contract,
                         refinement_harness, service_foundation, subscription_config
surfaced to planner    : the same, minus subscription_config

'config/subscriptions.yaml'    in planner body : False
'config/refinement_policy.yaml' in planner body : True
```

`refinement_harness` is the same shape — an app-level policy task with a null `capability_pack_id` and a narrow allowed path set — and it is surfaced with its config path. The only occurrence of "subscription" in the planner's entire injected context was the bare type name in the vocabulary line.

Then, against the real validator:

```
contract required_outputs: ['config/subscriptions.yaml']
what the run produced    -> REJECTED  (message identical to the live log)
the contract's own value -> ACCEPTED
```

**Not a regression.** The tuple was written 2026-05-07; the `subscription_config` contract landed 2026-06-14 and was never added to it. This has never worked — it surfaced now because #680 let a monetized greenfield plan reach admission for the first time.

## The change

One tuple entry. The data and the validator already agreed — the contract declares `required_outputs: [config/subscriptions.yaml]` and the validator demands exactly that. Only delivery to the planner was missing, so this adds **no new path, constant, or file**.

## Admission is untouched

This surfaces a contract; it does not relax a gate. Pinned by tests:

- a task owning the contract path passes admission
- a task owning nothing is **still rejected**, with the live message asserted
- four near-miss paths stay rejected: `app/config/subscriptions.yaml`, `config/subscriptions.yml`, an extra path alongside the correct one, and `modules/billing/contracts/subscriptions.yaml`

## The integrity guard

`test_every_task_contract_is_surfaced_to_the_planner` asserts **set equality** between `file_contracts.yaml` `task_contracts` and `_PLANNING_CONTRACT_ORDER`, deliberately in both directions:

- a contract missing from the tuple is invisible to the planner (this bug)
- a tuple entry with no matching data silently renders nothing (a typo would)

Excluding a contract from the planner on purpose should be a deliberate edit to that assertion, not a silent omission elsewhere. A parametrized companion asserts each contract's `required_outputs` actually appear in the body — surfacing the *name* is not enough, since the paths are what get copied.

`TestFileContractsIntegrity` already asserted these contracts exist in the data file. Nothing asserted they reach the agent told to copy from them, which is why a five-week-old gap surfaced in a live run instead of in CI.

## Deterministic repair: assessed, deferred

Not a small canonical derivation today, so per instruction it goes to a follow-up:

- `file_contracts.yaml` is read **only** by this prompt hook, so repairing in the review lane needs new asset-loading plumbing there
- the path exists as a bare literal in **four** places in `app_build_plan.py`, with **no shared constant**

Deriving from `required_outputs` would add a third source of truth. Doing it properly means extracting that constant first — a refactor of the validator, not an add-on to this fix. Worth doing: `_repair_plan`'s own docstring argues for exactly this case (*"Rejecting a plan over a name the system can derive... costs three LLM rounds and then the whole run"*), and `_repair_coverage` is module-scoped so an app-policy task with a null pack id falls outside it.

## Tests

`tests/test_planner_contract_surface.py` (new, 18 cases).

Verified non-vacuous by reverting the one-line fix and re-running:

```
without the fix:   3 failed, 15 passed
with the fix:     18 passed
```

The three that flip are the regression plus the integrity guard. The admission tests pass in **both** states — that is the point: they assert this change does nothing to ownership validation.

```
focused (contract / appgenerator / planner / file_contract /
  build_plan / plan_review / task_batch)   4487 passed, 22 skipped
ruff check .                               All checks passed!
source hygiene scan                        violations: 0
```

Full suite was still running locally when this opened; CI runs it here.

## Scope

Three files: the tuple entry, the new test, a CHANGELOG entry. App Zero untouched — its only AppGenerator build context contributes a hosted capability directory and nothing about file or task contracts, and `workflow_context_path` resolves strictly to the factory-owned build context, so a workspace cannot extend it.

## Not claimed

This gives the planner the contract it was told to copy from. It does not prove the model then copies it correctly — that needs a live run, deliberately held until this merges. The six already-surfaced contracts working is supporting evidence, not proof.


elCaptnCode:
The full local suite finished after this PR opened — clean:

```
pytest -q --no-cov     18521 passed, 118 skipped, 0 failed     (21:59)
```

18521 is exactly 18 above `origin/main`'s 18503, which accounts for this PR's 18 new cases and confirms nothing else moved.

That closes the one gap the description flagged ("full suite was still running locally when this opened").

PR #685, opened 2026-09-20
fix(app-generator): every typed companion manifest is an optional owned path

elCaptnCode:
## The defect

For a `module_contract` task that **owns** `contracts/events.yaml` while the module has **no events**, every output the worker could produce was rejected by something:

| Output shape | code_files guard | batch ownership | prompt |
|---|---|---|---|
| `events_yaml: null` + raw empty file *(what the live worker did)* | **REJECT** | — | forbidden |
| `events_yaml: null` + no file *(what the prompt mandates)* | OK | **REJECT** | ✓ |
| `events_yaml: {events: []}` typed-empty | OK | OK | forbidden ("not an empty placeholder") |

Live build `build_314a761d` at OSS `4769dcf7`: the auth module's worker set `events_yaml: null` as its prompt requires and, because it owned the path, also emitted the file raw. Rejected:

```
module_contract.events_yaml is null but raw output emits
modules/user_authentication/contracts/events.yaml.
```

The canonical output — null and no file, the shape the guard's own test describes as *"A module with no events declares null and emits nothing"* — was rejected too, reproduced offline from the preserved `rejected_output`:

```
did not emit required owned_paths: ['modules/user_authentication/contracts/events.yaml']
```

The only shape that passed both validators is the one the prompt forbids. The same worker built two zero-events modules in this run (`event_flows: []`); `task_management` happened to pick the typed-empty placeholder and lived, `user_authentication` followed its prompt and died. Which rule the model honored decided the build. `app.json` and both planned pages depend on every module contract for visibility, so one module with nothing to declare produced no app at all.

## Owner

`optional_task_output_paths(module_contract)` was a hand-written list of six companion manifests. The raw-file guard in `code_files.py` iterates `_MODULE_CONTRACT_OUTPUT_PATHS`, which names nine. The three missing — **`events.yaml`, `settings.yaml`, `admin.yaml`** — were therefore guarded as typed-only yet required as raw paths. `reactions.yaml` is the internal control: null and unemitted in the very same output, and it drew no complaint, because it *was* on the list.

Every other authority already called events.yaml optional: `file_contracts.yaml` (`optional_outputs`), the structured-output description (*"Use null unless…"*), the worker prompt, and the guard's test. Born this way in `a1cd0a16` (2026-06-01); #678 only renamed the function. Latent, not a regression.

## The change

**`task_batches.py`** — the `module_contract` optional set is now derived from `_MODULE_CONTRACT_OUTPUT_PATHS` minus `module_yaml`, so the raw-file guard and the ownership check name the same files by construction. Only `module.yaml` is required.

**`code_files.py`** — the guard's advice gains the zero-events branch. Before, a module with nothing to declare was told to *"Set module_contract.events_yaml to the contract"* — a field its prompt says to leave null — and the recovery retry reproduced the identical output until "repeated rejection is no progress" ended it (`attempts=2`, `recoverable=false`). It now says: set the field when the module declares one; when it is null, omit the file from `code_files` entirely, since a null field emits nothing.

No path is added, no placeholder created, no file is written that the worker did not declare through its typed field.

**`code_files.py`, second change — from adversarial review.** The guard fired only when the typed key was *present and null*; a bundle that *lacked* the key was silent, and making three more companions optional made that gap newly consequential. It now treats an absent key as null (`bundle.get(key) is None`). The strict structured output always carries every key, so no real output changes; only a hand-built payload could omit one, and a raw companion file must not slip past on that alone. Pinned by `test_an_absent_field_is_treated_as_null`, which fails at base.

## Adversarial review

Three lenses (downstream callers, weakened protections, test vacuity), every finding double-refuted by independent agents executing the scenarios against the worktree. Four claims raised, three survived, all rated *minor*:

- **batch ownership now accepts an unowned `events`/`settings`/`admin.yaml` under the task's own module** — the intended relaxation, identical to what the six other companions already had; content is still checked downstream (`module_loader` event/emits checks, admin-panel vs `admin_registry` validation). Acknowledged.
- **guard silent when the key is absent** — fixed above.
- **set-equality tests mirrored the implementation and were duplicated** — fixed: the guard now asserts an explicit literal of the nine paths *and* the derivation (a mirror alone cannot catch a key silently dropped from the map); the duplicate is removed.

The refuted claim — that repair admission could now write an unowned companion — was shown unreachable through the production repair path, with the same exposure already present at base. Five further findings came back `none`: the eight base failures target exactly the defect; the planned-completeness test is live, not vacuous; exclusivity, foreign-write protection, and the present-and-null guard are unchanged before and after.

## Not weakened — each proven by a test that passes in *both* states

Reverting the source and re-running the new tests: **8 fail at base, 56 pass after.** The tests that pass in both states are the point — they show these protections are unchanged rather than vacuously green:

- **exclusive task ownership** — a foreign task emitting the auth `events.yaml` is still rejected by `_validate_exclusive_task_paths`, and cannot claim it as its own optional path (`emitted files outside owned_paths`)
- **typed materialization** — a typed events manifest still renders through `_materialize_module_contract_file_map`, with the generator's property list compiled into a schema mapping (what a raw file would have skipped)
- **raw-file rejection** — null field + raw file is still refused with the same leading message
- **empty typed manifest** — `{events: []}` remains valid, so the shape the successful module used keeps working

Downstream consumers audited: `task_allowed_paths` is `(own | optional) − foreign`, so widening `optional` only ever adds paths under the task's **own** module id, and anything another task explicitly owns is still subtracted. `planned_artifact_diagnostics` subtracts the same function, so the batch gate and the completeness gate cannot diverge — pinned by a test.

## Tests

`tests/test_module_contract_owned_events_null.py` (new, 8) — the seven required regressions plus a second exclusivity variant. Fixture is the live task and output reduced to the fields the checks read.

`tests/test_task_batches_extended_helpers.py` (+2) — `events`/`settings`/`admin` are optional; and **set equality** with `_MODULE_CONTRACT_OUTPUT_PATHS − {module_yaml}`, asserted both as an explicit literal of the nine paths and as the derivation, so a key dropped from the guard's map fails here instead of silently shrinking both sets together.

`tests/test_contract_file_needs_its_typed_field.py` (+1 test, +3 assertions) — the message names the null case and what to do instead; an absent key is treated as null.

```
without the fix:   8 failed, 48 passed
with the fix:     56 passed
ruff check .       All checks passed!
source hygiene     violations: 0
```

Focused suite and full suite: see comments below.

## Not claimed

This gives a zero-events module a passing output and a retry message that points at it. It does not change the planner, which owned an optional file for a module the plan gave no events (its own rule says to add optional files "only when the product scope actually needs them"), and it does not change the drain: page bundles still fail on any failed module contract, including a visibility-only dependency. Both are flagged for separate PRs. No live acceptance has been run against this, deliberately, until it merges.


elCaptnCode:
Suites finished after this PR opened — both clean:

```
focused (task_batch / module_contract / code_files / task_integrity /
  contract_file / owned_path / repair_policy)        1027 passed, 16 skipped
full suite, final tree                               18532 passed, 118 skipped, 0 failed   (16:16)
```

Two numbers worth reading together. `origin/main` was 18521 after #682. A first full run on the pre-review tree gave 18532 (+11: nine tests in the new file, two in `extended_helpers`). The review-driven edits then removed one duplicate test and added one guard-hardening test, and the final-tree run is again 18532 — so the late edits changed exactly what they were meant to and nothing else in the suite moved.

That closes the "see comments below" note in the description. All local gates are now green against the exact commit under review (`ca6b7b26`).

PR #692, opened 2026-09-21
docs(contract): repository settings are not an agent's to change

elCaptnCode:
## Why

#691 gated the expensive paths with `CODEOWNERS`. The thing that actually went wrong is out of its reach: **branch protection is an API call, not a file**, so no owner entry can guard it. This rule is the only thing that can.

Every agent pushes as the same single collaborator, and GitHub does not let an author approve their own pull request. A nonzero required-approval count is therefore unsatisfiable by anyone — not slow, impossible. An agent that reads "blocked" as "lower the gate" will lower it.

That happened. An approval requirement was temporarily set to `0` to land a PR and restored afterwards, and a second agent's unrelated PR merged unreviewed inside that window. The work in both was sound and the change was disclosed; the mechanism was not the agent's to choose.

## What this adds

A `Repository Settings And Merge Authority` section, placed beside `Release Hold` because both cover actions that are cheap to take and expensive to undo:

- do not modify branch protection, rulesets, required checks, or repository settings
- do not merge with `--admin` or otherwise bypass a required check or a `CODEOWNERS` stop
- treat a blocked merge as a question for the maintainer, not an obstacle to route around

It also records *why* the `CODEOWNERS` paths cannot be approved under a single owner, so the next agent reads the stop as designed behavior rather than a defect to fix. Without that, the natural inference from "nobody can approve this" is "the gate is broken," which is how this started.

## Scope

One file, 27 added lines, no deletions. No code, no config, no behavior change.

## Checks

```
source hygiene    violations: 0
ruff check .      All checks passed!
```

`docs/` is not a `CODEOWNERS` path, so this PR also serves as a live check that ordinary work still merges on green CI under the new settings (`approvals=0`, `code_owner_reviews=true`, `enforce_admins=false`).

One test reads this file — `test_appgenerator_persistence_alignment.py` concatenates it with three other docs and asserts certain strings are **present**. This change is purely additive, so those assertions are unaffected. The filtered local run (`-k "contract or docs or governance or readiness"`) was still in flight when this opened; CI runs it across all eight shards regardless, and I'll post the local result below.


elCaptnCode:
The filtered local run finished — clean:

```
pytest -k "contract or docs or governance or readiness"
  3519 passed, 12 skipped, 0 failed   (3:21)
```

That covers `test_appgenerator_persistence_alignment.py`, the one test that reads `docs/agent-engineering-contract.md`. It concatenates that file with three other docs and asserts certain strings are present; this change is purely additive, so the assertions hold — now confirmed rather than reasoned.

Closes the "still in flight" note in the description. Local gates for this commit: hygiene 0, ruff clean, 3519 focused tests green.

PR #693, opened 2026-09-21
fix(app-registry): a draft start is not superseded by the build it just wrote

elCaptnCode:
## The defect

A monetized greenfield journey at OSS `859d6c59` stopped at the chat-binding gate, before any plan review:

```
POST /api/chats/mozaiks-platform/ValueEngine/start -> HTTP 500
ValueError: The selected build session has been superseded
```

No plan, no task batch, no artifacts, no validation. Earlier than any of the previous four runs.

`resolve_build_binding` mints a build id for the new run and persists it:

```python
binding = RunBuildBinding(build_id=f"build_{uuid4().hex}", phase="genesis")
result = await self.update_build_status(..., current_build_run={"build_id": binding.build_id, ...})
if not result["success"]:
    raise ValueError("Registered draft changed before its build could start")
```

then checks whether the registry moved underneath it:

```python
current_build_id = (record.get("current_build_run") or {}).get("build_id")
if (not refinement or resume) and current_build_id and current_build_id != binding.build_id:
    raise ValueError("The selected build session has been superseded")
```

`record` is the copy read **before** that write, so `current_build_id` is the build the run had just replaced. It can never equal the new binding, so the start always fails. The sibling create branch already does `record = result["app"]` for exactly this reason; this branch did not.

## Why it surfaced now, and why it is not a #690 regression

It only fires when the record already carried a `current_build_run` — a re-run against a prepared target, not a first build. App Zero's `prepare_greenfield_run.py` prepares a build id and clears the active chat, which is precisely that shape. The four earlier runs never reached this path.

`#690` never touched this file. The defect is latent and predates it.

## The change

One line: refresh `record` from the write result, mirroring the create branch.

**The race guard is intact.** The write is conditional on `expected_lifecycle_state`, so a concurrent writer makes it return `success=False` and raise `"Registered draft changed before its build could start"` three lines earlier. Reading back what was just stored removes a false positive, not a protection.

## Tests

`tests/test_factory_build_target.py` (+3), beside the existing supersession test:

- **the live case** — a draft that already carried a build starts cleanly, and the returned binding is the one actually persisted
- **a concurrent writer still stops the start** — the conditional write rejects it
- **reading back a write is not a way to create one** — without create authority the same call is still refused

Non-vacuity, by reverting the one line:

```
without the fix:   1 failed, 39 passed
with the fix:     40 passed
```

Exactly one test flips. The other two pass in both states, which is the point — they assert this change does not widen anything.

```
focused (app_registry / build_target / build_binding /
  launcher / session / registry / lifecycle)   782 passed, 14 skipped
ruff check .                                   All checks passed!
source hygiene                                 violations: 0
```

Full suite running; result posted below.

## Known, not fixed here

The same run showed a second defect: after the 500, the websocket resolved to `draft-app-023c1075` instead of the requested `release-greenfield-value-target`. `allow_create` is keyed on workflow name alone (`build_target.py`: `allow_create=workflow_name in {"ValueEngine", "ExistingAppDiscovery"}`), so a retry that has lost its `build_registry_id` falls into the create branch and silently binds a different app.

That is arguably the more dangerous of the two — a run that quietly builds into the wrong target produces plausible evidence about the wrong thing. It gets its own PR. It is not bundled here because this fix removes the 500 that triggers it, and the blocker should ship without waiting on a broader fail-closed change.

## Not claimed

This clears the chat-binding gate. It says nothing about the gates behind it, which no run has yet reached. Live acceptance is deliberately held until this merges.


elCaptnCode:
Confirmed the exact code path from the run's own evidence, rather than by reading alone.

`greenfield-setup.json` records the state `prepare_greenfield_run.py` left the target in:

```
old_active_chat_id    : e90afc26-8383-447a-99f6-746b1bfca732
new_active_chat_id    : None
new_lifecycle_state   : draft
new_current_build_run : {'build_id': 'build_4c0af385c27840289a09994f7485a7a6', ...}
```

`active_chat_id` cleared, `lifecycle_state` draft, and a build id already on the record. At chat start that routes into exactly the branch this PR fixes:

- `build_registry_id` supplied, `record["active_chat_id"]` falsy → `if not active_chat:`
- `lifecycle_state == "draft"` and `allow_create=True` (ValueEngine) → mint a new build id, persist it
- final check reads the **pre-write** `record`, so `current_build_id` is `build_4c0af385…`
- compared against the freshly minted id → mismatch → `"The selected build session has been superseded"`

Which matches `journey-gate-failure.json` exactly: `prepared_build_id: build_4c0af385c27840289a09994f7485a7a6`, `chat_start_http_status: 500`.

Worth noting the other route to the same message, which this PR deliberately does **not** change: had `active_chat_id` still been set, the else branch would have resolved the *old* chat's binding and the check would have rejected it against the newly prepared build. That rejection is correct — a stale chat's binding genuinely is superseded. Only the read-your-own-write case was wrong, and only that case is changed.

One consequence to watch in the next run: the prepared build id is discarded and OSS mints its own, so `prepared_build_id` in the harness evidence will not match the build the run actually executes. That is pre-existing behaviour, not introduced here, but any harness assertion comparing the two will need to compare against the binding OSS returns.

